PR #33, opened 2015-04-28
Add no-default-features, providing just the parsing functionality

pyfisch:
It is not practical for some packages to pull a dependency on openssl,
but they want to use cookie parsing, so compiling with
--no-default-features removes the cookie jar and with this the dependency
on openssl.

As discussed in hyperium/hyper#433


alexcrichton:
Thanks!

PR #34, opened 2015-05-28
include jar even if secure feature is disabled

seanmonstar:
The `secure` module and the `signed` and `encrypted` methods are the
only parts of `jar` that depend on `openssl`, so this adds all the other
parts of jar to be used even without those crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,8 @@ similar to Rails' cookie jar.
 """
 
 [features]
-default = ["openssl", "rustc-serialize"]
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Looks like there's some travis failures though in the doc tests.

seanmonstar:
Hm, how does `cargo doc --no-deps` work?

seanmonstar:
Ok, I moved the docs using those methods to the specific methods themselves. Passes locally.

PR #37, opened 2015-07-27
Use base64 instead of hex for encoding

Ryman:
This should be more compact in general and allow more data to be stored in secure cookies.

Technically, this is a breaking change as it invalidates cookies from previous versions.


alexcrichton:
Thanks!

PR #39, opened 2015-08-11
Allow parsing of negative Max-Age attribute-values

pfernie:
Per [RFC6265 Section 5.2.2](http://tools.ietf.org/html/rfc6265#section-5.2.2), Max-Age may be negative or 0, which a server uses to expire a cookie. First parse `max_age` as a signed value, and then clamp the result to the positive range (as negative values have the same meaning as `0`: expire the cookie). This limits the maximum parseable value to `i64::max_value()` (from `u64::max_value()`), but such large Max-Age values seem meaningless.


alexcrichton:
Thanks!

PR #41, opened 2015-11-20
chore(cargo): update url to 0.5

untitaker:
~~I also allow 0.2 because it still works with that version~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ default = ["secure"]
 secure = ["openssl", "rustc-serialize"]
 
 [dependencies]
-url = "0.2"
+url = "0.5"
 time = "0.1"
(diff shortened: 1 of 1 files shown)


untitaker:
I see, would you accept abandoning Rust 1.0 support in favor of just 1.4?

alexcrichton:
Is there a particularly pressing reason to do this? This isn't a "public dependency" of this crate (e.g. not a shared type), so it should be fine to leave this as-is

untitaker:
No, no particular hurry. Just thought this might be useful because of binary
size.

untitaker:
Updated PR to just pin to latest version.

alexcrichton:
Thanks!

SimonSapin:
Could you get this on crates.io? Thanks!

PR #42, opened 2016-03-02
Update to rust-url 1.0

SimonSapin:
**Do not merge yet:** rust-url 1.0 is not published yet and may still get breaking changes. The goal of this PR for now is to demonstrate API usage.

Depends on https://github.com/servo/rust-url/pull/176

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ default = ["secure"]
 secure = ["openssl", "rustc-serialize"]
 
 [dependencies]
-url = "0.5"
+url = "1.0"
 time = "0.1"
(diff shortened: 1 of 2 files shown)


SimonSapin:
https://crates.io/crates/url/1.0.0 is (finally!) published, so this is ready to land. Could you also publish it on crates.io?

alexcrichton:
Nice!

SimonSapin:
It realized that this crates only uses percent-encoding. Do you think it would be useful to extract that from the rest of `url` into a separate crate?

alexcrichton:
Perhaps? I'd be fine just leaving the dependency on `url`, not a burning desire of me to slim that down at all

PR #44, opened 2016-04-23
Add RuctcEncdoable and RustcDecodable to derive attribute for Cookie

DDEFISHER:
It was recommended in this pull [request](https://github.com/servo/servo/pull/10800) for servo that I placed the derive attribute for the RustcEncodable and RustcDecodable traits for the cookie_rs:Cookie to avoid encoding and decoding manually.  I noticed in the time library that the  #[cfg] attribute is used to make the encodable/decodable attribute optional.  If you would like me to add the #[cfg] attribute just let me know!

Edit: Actually I went ahead and added the #[cfg] attribute because rustc-serialize is marked as optional in Cargo.toml

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ similar to Rails' cookie jar.
 
 [features]
 default = ["secure"]
+serialize-rustc = ["rustc-serialize", "time/rustc-serialize"]
 secure = ["openssl", "rustc-serialize"]
 
 [dependencies]
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ script:
   - cargo build --verbose
   - cargo test --verbose --no-default-features
   - cargo test --verbose
+  - cargo test --verbose --features serialize-rustc
   - rustdoc --test README.md -L target
   - cargo doc --no-deps
 after_success: |
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,7 @@
 
 extern crate url;
 extern crate time;
+#[cfg(feature = "serialize-rustc")] extern crate rustc_serialize;  
 
 use std::ascii::AsciiExt;
 use std::collections::BTreeMap;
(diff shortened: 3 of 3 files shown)


frewsxcv:
Maybe update `.travis.yml` to run `cargo test --features "rust-serialize"`

frewsxcv:
Nevermind my comment, didn't realize it's a `default`  feature.

Cargo.toml
```diff
@@ -18,6 +18,6 @@ secure = ["openssl", "rustc-serialize"]
 
 [dependencies]
 url = "1.0"
-time = "0.1"
+time = { version = "0.1", features = ["rustc-serialize"] }
```
alexcrichton: This seems a little unfortunate that it unconditionally enables rustc-serialize in time. Could this perhaps use a different feature name which enables both the rustc-serialize crate as well as the rustc-serialize feature in `time`?

DDEFISHER: I was thinking there had to be a better way.  I will try your suggestion.

DDEFISHER:
I think this is better, I was not sure if I should include secure in the feature or have feature "secure" and "serialize" contain rustc-serialize.  I also added a line for the serialize feature to travis.yml to address the previous suggestion of @frewsxcv .

alexcrichton:
Thanks! On second thought, could you name the feature `serialize-rustc`? That leaves the door open to add `serialize-serde` in the future.

Also for now let's have `secure` include `rustc-serialize` as it shouldn't also enable the `RustcSerialize` impls (even though the dependency is available). Right now we just need base64 which should in theory come from something other than `rustc-serialize` and otherwise that'd be a breaking change.

DDEFISHER:
Changed feature name to serialize-rustc

alexcrichton:
Thanks!

Ms2ger:
@alexcrichton can we get a crates.io release?

frewsxcv:
Opened a PR to bump the version: https://github.com/alexcrichton/cookie-rs/pull/45

PR #47, opened 2016-05-16
Change the way to parse cookie attributes

### LINKED ISSUE #46: Incorrectly parse "secure" attributes from cookies
From http://tools.ietf.org/html/rfc6265#section-5.2 the secure attribute may have a value (which must be ignored).

This means that a cookie like `foo=bar ;HttpOnly; Secure=aaaa` should set the secure flag of cookie to true, which is not the case right now.

I catched this bug by running the cookie test suite from https://github.com/abarth/http-state (see tests test_attribute000[4578]) which is focused on the secure flag. But the same goes for other attributes like httponly.

I will propose a fix soon.

fduraffourg:
Fix #46 

To be totally RFC6265 compliant (see section 5.2 item 4), attr_split should return two strings whatever the case. If the value is not specified it should be "". But that would break the test cases (and potentially code using this crate) as some attributes would be added to `custom` (like the "Ignored" attribute from the test case).

So I used an option for the value which allow the match case to choose what to do depending of whether the value is set or not.

This fix the parsing of `httponly` and `secure` without breaking the current behavior.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,59 +65,66 @@ impl Cookie {
         c.value = try!(percent_decode(value));
 
         for attr in pairs {
-            let trimmed = attr.trim();
-            match &trimmed.to_ascii_lowercase()[..] {
-                "secure" => c.secure = true,
-                "httponly" => c.httponly = true,
-                _ => {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #48, opened 2016-05-29
Implement serde serialization for Cookie

danlrobertson:
Implement `serde::Serialize` and `serde::Deserialize` for `Cookie` and add it behind the `serialize-serde` feature. This is largely done to allow for `Cookie` to be sent across servo `ipc-channel`s.

Critiques are very welcome. There might be a better way of doing this.

NB: `Cookie::custom` was not included in this PR.

Related to: servo/servo#10826

CC @asajeffrey

---
This change is [Reviewable](https://reviewable.io/reviews/alexcrichton/cookie-rs/48)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,9 +16,14 @@ similar to Rails' cookie jar.
 default = ["secure"]
 serialize-rustc = ["rustc-serialize", "time/rustc-serialize"]
 secure = ["openssl", "rustc-serialize"]
+serialize-serde = ["serde"]
 
 [dependencies]
 url = "1.0"
 time = "0.1"
 rustc-serialize = { version = "0.3", optional = true }
 openssl = { version = "0.7.0", optional = true }
+serde = { version = "0.7.5", optional = true }
+
+[dev-dependencies]
+serde_json = "0.7.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,11 +4,14 @@
 extern crate url;
 extern crate time;
 #[cfg(feature = "serialize-rustc")] extern crate rustc_serialize;  
+#[cfg(feature = "serialize-serde")] extern crate serde;
 
 use std::ascii::AsciiExt;
 use std::collections::BTreeMap;
 use std::fmt;
 use std::str::FromStr;
+
+#[cfg(feature = "serialize-serde")] use serde::{Serialize, Deserialize};
 
 pub use jar::CookieJar;
 mod jar;
@@ -142,6 +145,40 @@ impl Cookie {
     }
 }
 
+#[cfg(feature = "serialize-serde")]
+impl Serialize for Cookie {
+    fn serialize<S>(&self, serializer: &mut S) -> Result<(), S::Error>
+        where S: serde::Serializer
+    {
+        serializer.serialize_str(&*self.to_string())
+    }
+}
+
+#[cfg(feature = "serialize-serde")]
+impl Deserialize for Cookie {
+    fn deserialize<D>(deserializer: &mut D) -> Result<Cookie, D::Error>
+        where D: serde::Deserializer
+    {
+        deserializer.deserialize_string(CookieVisitor)
+    }
+}
+#[cfg(feature = "serialize-serde")]
+struct CookieVisitor;
+
+#[cfg(feature = "serialize-serde")]
+impl serde::de::Visitor for CookieVisitor {
+    type Value = Cookie;
+
+    fn visit_str<E>(&mut self, v: &str) -> Result<Cookie, E>
+        where E: serde::de::Error
+    {
+        match Cookie::parse(v) {
+            Ok(cookie) => Ok(cookie),
+            Err(()) => Err(serde::de::Error::custom("Could not parse serialized cookie!"))
+        }
+    }
+}
+
 pub struct AttrVal<'a>(pub &'a str, pub &'a str);
 
 impl<'a> fmt::Display for AttrVal<'a> {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 
 name = "cookie"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
-version = "0.2.4"
+version = "0.2.5"
```
alexcrichton: Could you back this out? I'll do a separate commit once tests pass

Cargo.toml
```diff
@@ -16,9 +16,18 @@ similar to Rails' cookie jar.
...
 
 [dependencies]
 url = "1.0"
 time = "0.1"
 rustc-serialize = { version = "0.3", optional = true }
 openssl = { version = "0.7.0", optional = true }
+
+[dependencies.serde]
+version = "0.7.5"
+optional = true
```
alexcrichton: Could this use the same syntax as openssl/rustc-serialize above?

Cargo.toml
```diff
@@ -16,9 +16,18 @@ similar to Rails' cookie jar.
...
+
+[dependencies.serde_json]
```
alexcrichton: This dependency isn't needed, right?

src/lib.rs
```diff
@@ -142,6 +145,197 @@ impl Cookie {
...
+             expires: expires,
+             max_age: max_age,
+             domain: domain,
+             path: path,
+             secure: secure,
+             httponly: http,
+             custom: BTreeMap::new()
+         })
+     }
+}
```
alexcrichton: Could this actually go into a serde-specific module? That'd help cut down on the number of `#[cfg]` directives.

Cargo.toml
alexcrichton: Oh this may want to just be a `dev-dependencies` entry

alexcrichton:
Thanks @danlrobertson! Could you also edit the CI configuration to run these tests as well?

Cargo.toml
danlrobertson: Np, I can do that tomorrow morning.

danlrobertson:
Updated

src/lib.rs
```diff
@@ -190,6 +212,178 @@ impl FromStr for Cookie {
     }
 }
 
+#[cfg(feature = "serialize-serde")]
+mod ser {
```
danlrobertson: Debated putting in a separate file, but since it is only used here, I thought it might be simpler to leave it here.

src/lib.rs
```diff
@@ -190,6 +212,178 @@ impl FromStr for Cookie {
...
+                 None => try!(v.missing_field("name"))
+             };
+
+             let value = match value {
+                 Some(value) => value,
+                 None => try!(v.missing_field("value"))
+             };
+
+             let expires = expires.map(|expires| {
+                 time::strptime(&*expires, "%a, %d %b %Y %H:%M:%S %Z").unwrap()
```
alexcrichton: Could this `unwrap()` be translated to an error?

src/lib.rs
```diff
@@ -190,6 +212,178 @@ impl FromStr for Cookie {
...
+             });
+
+             let secure = match secure {
+                 Some(secure) => secure,
+                 None => try!(v.missing_field("secure"))
+             };
+
+             let http = match http {
+                 Some(http) => http,
+                 None => try!(v.missing_field("secure"))
```
alexcrichton: s/secure/http/

src/lib.rs
```diff
@@ -190,6 +212,178 @@ impl FromStr for Cookie {
...
+                value: &*cookie.value,
+                expires: &cookie.expires,
+                max_age: &cookie.max_age,
+                domain: &cookie.domain,
+                path: &cookie.path,
+                secure: cookie.secure,
+                httponly: cookie.httponly
+            }
+        }
+    }
```
alexcrichton: Shouldn't `custom` also be serialized here? Also, would it perhaps be easier to just serialize the stringified version of a cookie?

danlrobertson: Yes it should. That would probably be easier. Let me know if you'd like me to try that.

src/lib.rs
danlrobertson: good catch

src/lib.rs
danlrobertson: I don't think so. We explicitly format the date like that... I could be wrong, but again I don't think so

alexcrichton: Ah yeah this'll always succeed for data created from serialization, but it may be possible that something like this could change over time or perhaps the data is user generated. In that sense it should propagate the error.

This may become moot though if serialization just goes through a string

src/lib.rs
alexcrichton: Yeah let's just stick to using a string for now as that should be easiest for manually implementing the trait.

danlrobertson:
After I thought about it more, I realized it would be increadibly easy to add `custom` to the current code. I also realized that converting the cookie to a string and then back would not lead to "true json" when serialized. If you think the extra code is worth having serialized json then I'd suggest keeping the current code, but if you want a smaller code base, I have commit using cookies string methods.

alexcrichton:
Yeah let's stick to a string for now as it should be a lossless way to interpret a cookie as that's where it came from anyway (also a more compressed form of a serialization format)

danlrobertson:
Updated. Good point.

src/lib.rs
```diff
@@ -142,6 +145,49 @@ impl Cookie {
...
+#[cfg(feature = "serialize-serde")]
+impl serde::de::Visitor for CookieVisitor {
+    type Value = Cookie;
+
+    fn visit_str<E>(&mut self, v: &str) -> Result<Cookie, E>
+        where E: serde::de::Error
+    {
+        match Cookie::parse(v) {
+            Ok(cookie) => Ok(cookie),
+            Err(_) => Err(serde::de::Error::custom("Could not parse serialized cookie!"))
```
alexcrichton: Could this pass along the error instead of swallowing it up?

src/lib.rs
```diff
@@ -142,6 +145,49 @@ impl Cookie {
...
+    }
+
+    fn visit_string<E>(&mut self, v: String) -> Result<Cookie, E>
+        where E: serde::de::Error
+    {
+        match Cookie::parse(&*v) {
+            Ok(cookie) => Ok(cookie),
+            Err(_) => Err(serde::de::Error::custom("Could not parse serialized cookie!"))
+        }
+    }
```
alexcrichton: I think this is a default method delegating to `visit_str` by default, so perhaps this could be omitted?

src/lib.rs
danlrobertson: It's a `Err(())`. Is there any usefull iformation in that? I may be mistaken on this though. I'll double check later.

src/lib.rs
danlrobertson: Good point

src/lib.rs
alexcrichton: Oh oops, right yeah forgot about that. In that case though could you change this to `Err(())` as a pattern match so we can be sure to update it if the error type changes?

danlrobertson: Ah, good point... Updated

alexcrichton:
Thanks @danlrobertson!

PR #52, opened 2016-07-30
Allow serde 0.8

nox:


alexcrichton:
Thanks @nox! I'm considering making a major version bump instead of supporting both versions here though because Cargo doesn't handle these sorts of version dependencies that well. Unfortunately that would mean that a ton of crates would need to get bumped which may just be unnecessary pain, so I'm still thinking about the best solution here...

nox:
According to crates.io, only hyper relies on the serde feature of this crate. Given this fact I'm pretty sure we can use a range.

nox:
In fact, not even hyper enables the `serde` feature through its own `serde-serialization` feature. I'm confident we can go ahead with this PR.

alexcrichton:
Hm ok I'm going to take the route of releasing a new major version to ensure that over time if we pick up more serde features which aren't compatible with 0.7 we can continue doing so. Thanks for the PR though!

PR #54, opened 2016-09-03
Bump url to 1.2 (no changes)

spk:
Just updating url crates
Cheers


alexcrichton:
Thanks for the PR! Due to the way dependencies work in Cargo, however, this actually isn't necessary. Right now this is just saying that this crate is compatible with the 1.0 release of url, whereas with this change it would up the minimum to 1.0 (which in this case isn't necessary).

Cargo will automatically assign 1.2 as a dependency if available, though!

Issue #1, opened 2014-11-07
Cookie::parse could return an Option


seanmonstar:
Since the Err case returns unit, it may as well be an Option?

spk:
Can be closed now that #55 is merged ?

alexcrichton:
Indeed!

PR #55, opened 2016-09-03
Add CookieError for error handling

spk:
see https://github.com/alexcrichton/cookie-rs/issues/1
Cheers

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,15 +31,20 @@ pub struct Cookie {
 }
 
 #[derive(Debug)]
-pub enum CookieError {
-    ParseError,
-    Utf8Error(std::str::Utf8Error),
+pub struct Error {
+    inner: ErrorInner,
 }
 
-fn percent_decode(input: &str) -> Result<String, CookieError> {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! Could this actually be called `ParseError` as it's returned from `Cookie::parse`. Additionally could it be made into an opaque struct so it can be made extensible in the future as well?

spk:
Glad to help, there is also `percent_decode` that return `std::str::Utf8Error`
I don't know about "opaque struct" I will look it up, thanks

alexcrichton:
Oh sorry by "opaque struct" I mean something like:

```rust
pub struct Error {
    // ... private internals
}
```

vs

```rust
pub enum Error {
    // public variants
}
```

The difference is that with a public enum adding a variant is a breaking change, but with an opaque struct adding fields is not an API-breaking-change.

spk:
Thanks for the explanation ! I've made some changes in 3dcbde55d349010c85b90cc3aee6984820cf3a26

alexcrichton:
Thanks!

PR #61, opened 2016-10-28
HBump to 0.3.1

pwoolcoc:
Adds the `#[deny(missing_docs)]` attribute and add docs to get it to
compile


alexcrichton:
Thanks! Could you back out the bump to 0.3.2? That shouldn't be necessary to regenerate docs and I personally prefer to do that as a separate commit as well.

pwoolcoc:
Sorry about that! I changed the version back

alexcrichton:
Thanks!

PR #63, opened 2016-11-06
Sign entire cookie

### LINKED ISSUE #62: Feature: Sign entire cookie, not just value
I'd like signatures on cookies to include expiration date and path as well. The signature could still be inside the cookie value.

I would love to prepare a PR for this, but I'm unsure how to design this in a backwards compat way (or if that is even necessary).

untitaker:
Fix #62

### DIFF
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -387,7 +387,7 @@ mod secure {
     // https://github.com/rails/rails/blob/master/activesupport/lib
     //                   /active_support/message_verifier.rb#L70
     pub fn sign(key: &[u8], mut cookie: Cookie) -> Cookie {
-        let signature = dosign(key, &cookie.value);
+        let signature = dosign(key, &cookie.to_string());
         cookie.value.push_str("--");
         cookie.value.push_str(&signature.to_base64(STANDARD));
         cookie
@@ -414,7 +414,7 @@ mod secure {
             let (text, signature) = match split_value(&cookie.value) {
                 Some(pair) => pair, None => return None
             };
-            let expected = dosign(key, text);
+            let expected = dosign(key, &cookie.to_string());
             if expected.len() != signature.len() ||
                !memcmp::eq(&expected, &signature) {
                 return None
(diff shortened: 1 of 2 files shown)


src/jar.rs
```diff
@@ -414,7 +414,7 @@ mod secure {
             let (text, signature) = match split_value(&cookie.value) {
                 Some(pair) => pair, None => return None
             };
-            let expected = dosign(key, text);
+            let expected = dosign(key, &cookie.to_string());
```
alexcrichton: Presumably this shouldn't work, right? We should be verifying the signature of `text` here but instead it's verifying a signature of a signature?

untitaker: You're right, I wonder why the tests pass though.

src/jar.rs
```diff
@@ -540,17 +540,24 @@ mod test {
...
+            assert!($c.$secure().find("test").unwrap().value == "test");
+            let mut cookie = $c.find("test").unwrap();
+            cookie.expires = Some(cookie.expires.unwrap_or(0) + 1);
+            $c.add(cookie);
+            assert!($c.$secure().find("test").is_none());
         })
     }
 
-    #[cfg(features = "secure")]
+    #[cfg(feature = "secure")]
```
untitaker: This is so fucked up. Do you think I should file an issue against rustc for this?

src/jar.rs
untitaker: Fixed!

untitaker:
@alexcrichton ping

alexcrichton:
Ah sorry haven't forgotten about this, just haven't had a chance to think through the backcompat story here yet

untitaker:
@alexcrichton Bump. IMO it's not that big of a deal to invalidate all signed/encrypted cookies, as the browser doesn't guarantee either that those are stored. This is just session/cache expiration after all.

alexcrichton:
Yeah that's true, let's just merge.

untitaker:
Could you release a new version as well? Thanks!

alexcrichton:
Sure, published 0.4.0

alexcrichton:
Wait @untitaker this ended up breaking crates.io and now I'm confused. How does this work? Browsers don't send cookie parameters, just the values back. How is this supposed to be used?

untitaker:
Ack, feel free to revert

untitaker:
Sorry. I didn't test it locally, and assumed that the entire cookie including params would be sent by the client.

The idea (and my usecase) would be to securely expire a cookie without encoding the expiration date twice.

Issue #40, opened 2015-11-18
Why the percent-decoding on key & value?


badboy:
I hit a problem with cookie-rs as used by Servo.
A full and unnecessary long explanation [is online](http://badboy.github.io/cookie-bug/cookie_bug/index.html).

It comes down to this: cookie-rs percent-decodes key and value when parsing and stores them as Strings.
Servo uses these without modification to send back as a HTTP header.

This breaks if the cookie is not to be interpreted as a percent-encoded value (e.g. having newlines `%0A` or things that make it invalid UTF-8: `%EE`).
The actual decoding [dates back to last year](https://github.com/alexcrichton/cookie-rs/commit/724003decad72cdbe3f998b0b8d181682e9582c5).
It's undocumented and suprising IMO.

Any insights on why it was done this way?

alexcrichton:
Holy cow, that is indeed quite thorough! I think I did this at some point for crates.io, although I forget now the exact reason as to why. I'd be ok going back to just not auto-decoding it!

This crate was written pretty quickly to get crates.io up and running and wasn't exactly designed with the best "cookie practices" in mind, so I have a feeling it has room to improve in a few areas.

retep998:
I ran into this issue when working with mediawiki API, one of the cookies I was getting had some percent encoded symbols, which this library then decoded and never re-encoded when sending them back since they weren't part of the default encode set, which resulted in the cookie not having the exact same value and everything not working. So I'm currently using a local override of cookie-rs that has the encoding/decoding completely stripped out. https://github.com/retep998/cookie-rs/commit/0d19cc1c1741f326870fa09eeceb24f2bbe08f9d

jdm:
@retep998's patch makes signing into Slack work on Servo, too.

SergioBenitez:
@alexcrichton This issue's been open a little over a year now. It seems that there's a solution. Any chance of incorporating it or providing feedback as to why it might not be the right one?

alexcrichton:
@SergioBenitez of course! Unfortunately I don't have a lot of time to create a patch myself, but I don't mind reviewing PRs!

PR #66, opened 2016-12-19
Get rid of url encoding in cookies.

retep998:
Fixes https://github.com/alexcrichton/cookie-rs/issues/40


Ms2ger:
Can we get this on crates.io?

retep998:
Because of the way that this changes behavior in a likely backwards incompatible manner, it should probably be published as a new major version.

PR #67, opened 2016-12-29
Bump version to 0.4.1.

jdm:
As far as I can tell 9e4baf40ebe59b025b7e9a63b50dcc0460dc78a5 avoided any issues with breaking changes, so a minor version bump should be acceptable.


alexcrichton:
Unfortunately it's semantically a large breaking change, so I'll make a major release bump.

Issue #50, opened 2016-06-09
Use an AEAD cipher instead of AES-CBC + HMAC-SHA1


DemiMarie:
cookie-rs currently uses AES-256-CBC and then HMAC-SHA1 in encrypt-then-MAC.  This is secure, since AES-256-CBC and HMAC-SHA1 are both individually secure and because they are used in encrypt-then-MAC.

However, neither AES-256-CBC nor HMAC-SHA1 is fast, even when AES is hardware accelerated (due to the serial nature of CBC mode).  ChaCha20-Poly1305 is much faster.

Again, cookie-rs's current crypto is not _insecure_, just slow.

alexcrichton:
Ideally this'd actually be configurable rather than only hardcoding one and calling it a day, but I'm somewhat hesitant to deviate away from the source of this, Rails, which is probably much more experienced than I :)

frewsxcv:
Opened a PR: https://github.com/alexcrichton/cookie-rs/pull/68.

briansmith:
There's another PR for this: #76.

PR #68, opened 2017-01-08
Migrate from rust-openssl to *ring* for 'secure' functionality.

frewsxcv:
This also changes the encrypted' strategy from AES 256 CBC with an HMAC to AEAD ChaCha20 Poly1305.

https://github.com/alexcrichton/cookie-rs/issues/50

[breaking-change]

### DIFF
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -378,7 +378,7 @@ mod secure {
     extern crate rustc_serialize;
 
     use Cookie;
-    use self::ring::{aead, constant_time, digest, hmac, rand, pbkdf2};
+    use self::ring::{aead, digest, hmac, rand, pbkdf2};
     use self::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
 
     /// Algorithm used to sign the cookie value
@@ -401,7 +401,8 @@ mod secure {
 
     pub fn sign(key: &[u8], mut cookie: Cookie) -> Cookie {
         assert_eq!(key.len(), SIGNING_KEY_LEN);
-        let signature = dosign(key, &cookie.value);
+        let signing_key = hmac::SigningKey::new(SIGNING_ALGORITHM, key);
+        let signature = hmac::sign(&signing_key, cookie.value.as_bytes());
         cookie.value.push_str(SIGNATURE_SEPARATOR);
         cookie.value.push_str(&signature.as_ref().to_base64(STANDARD));
         cookie
@@ -429,19 +430,16 @@ mod secure {
         let (text, signature) = match split_value(&signed_value) {
             Some(pair) => pair, None => return None
         };
-        cookie.value = text.to_owned();
-
-        let expected = dosign(key, text);
-        if constant_time::verify_slices_are_equal(expected.as_ref(),
-                                                  &signature).is_err() {
-            return None
+        let verification_key =
+            hmac::VerificationKey::new(SIGNING_ALGORITHM, key);
+        let is_valid_signature = hmac::verify(
+            &verification_key, text.as_bytes(), &signature).is_ok();
+        if is_valid_signature {
+            cookie.value = text.to_owned();
+            Some(cookie)
+        } else {
+            None
         }
-        Some(cookie)
-    }
-
-    fn dosign(key: &[u8], val: &str) -> digest::Digest {
-        let signing_key = hmac::SigningKey::new(SIGNING_ALGORITHM, key);
-        hmac::sign(&signing_key, val.as_bytes())
     }
 
     pub fn encrypt_and_sign(key: &[u8], mut cookie: Cookie) -> Cookie {
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -23,7 +23,6 @@ use Cookie;
 /// # Example
 ///
 /// ```
-/// # #![allow(unstable)]
 /// use cookie::{Cookie, CookieJar};
 ///
 /// let c = CookieJar::new(b"f8f9eaf1ecdedff5e5b749c58115441e");
@@ -54,23 +53,23 @@ type Read = fn(&Root, Cookie) -> Option<Cookie>;
 type Write = fn(&Root, Cookie) -> Cookie;
 
 #[cfg(feature = "secure")]
-type SigningKey = Vec<u8>;
+struct SecureKeys {
+    key256: [u8; 256 / 8],
+    key512: [u8; 512 / 8],
+}
 #[cfg(not(feature = "secure"))]
-type SigningKey = ();
+type SecureKeys = ();
 
 #[cfg(feature = "secure")]
-fn prepare_key(key: &[u8]) -> Vec<u8> {
-    if key.len() >= secure::MIN_KEY_LEN {
-        key.to_vec()
-    } else {
-        // Using a SHA-256 hash to normalize key as Rails suggests.
-        // See https://github.com/rails/rails/blob/master/activesupport/lib/active_support/message_encryptor.rb
-        secure::prepare_key(key)
+fn prepare_keys(secret: &[u8]) -> SecureKeys {
+    let (key256, key512) = secure::generate_keys(secret);
+    SecureKeys {
+        key256: key256,
+        key512: key512,
     }
 }
-
 #[cfg(not(feature = "secure"))]
-fn prepare_key(_key: &[u8]) -> () {
+fn prepare_keys(_secret: &[u8]) -> () {
     ()
 }
 
@@ -78,7 +77,7 @@ struct Root {
     map: RefCell<HashMap<String, Cookie>>,
     new_cookies: RefCell<HashSet<String>>,
     removed_cookies: RefCell<HashSet<String>>,
-    _key: SigningKey,
+    keys: SecureKeys,
 }
 
 /// Iterator over the cookies in a cookie jar
@@ -88,16 +87,17 @@ pub struct Iter<'a> {
 }
 
 impl<'a> CookieJar<'a> {
-    /// Creates a new empty cookie jar with the given signing key.
+    /// Creates a new empty cookie jar with the given secret.
     ///
-    /// The given key is used to sign cookies in the signed cookie jar.
-    pub fn new(key: &[u8]) -> CookieJar<'static> {
+    /// The given secret is used to generate keys which are used to sign
+    /// cookies in the signed cookie jar.
+    pub fn new(secret: &[u8]) -> CookieJar<'static> {
         CookieJar {
             flavor: Flavor::Root(Root {
                 map: RefCell::new(HashMap::new()),
                 new_cookies: RefCell::new(HashSet::new()),
                 removed_cookies: RefCell::new(HashSet::new()),
-                _key: prepare_key(key),
+                keys: prepare_keys(secret),
             })
         }
     }
@@ -209,10 +209,10 @@ impl<'a> CookieJar<'a> {
         };
 
         fn design(root: &Root, cookie: Cookie) -> Option<Cookie> {
-            secure::design(&root._key, cookie)
+            secure::design(&root.keys.key512, cookie)
         }
         fn sign(root: &Root, cookie: Cookie) -> Cookie {
-            secure::sign(&root._key, cookie)
+            secure::sign(&root.keys.key512, cookie)
         }
     }
 
@@ -246,10 +246,10 @@ impl<'a> CookieJar<'a> {
             })
         };
         fn read(root: &Root, cookie: Cookie) -> Option<Cookie> {
-            secure::design_and_decrypt(&root._key, cookie)
+            secure::design_and_decrypt(&root.keys.key256, cookie).ok()
         }
         fn write(root: &Root, cookie: Cookie) -> Cookie {
-            secure::encrypt_and_sign(&root._key, cookie)
+            secure::encrypt_and_sign(&root.keys.key256, cookie)
         }
     }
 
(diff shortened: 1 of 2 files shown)


frewsxcv:
cc @SergioBenitez

src/jar.rs
```diff
@@ -410,95 +419,111 @@ mod secure {
...
         }
         Some(cookie)
     }
 
-    fn dosign(key: &[u8], val: &str) -> Vec<u8> {
-        let pkey = PKey::hmac(key).unwrap();
-        let mut signer = Signer::new(MessageDigest::sha1(), &pkey).unwrap();
-        signer.update(val.as_bytes()).unwrap();
-        signer.finish().unwrap()
+    fn dosign(key: &[u8], val: &str) -> digest::Digest {
```
Philipp91: nit: If you eliminate dosign(), you can replace the second of its only two usages with `verify()` instead of another `sign()`
https://briansmith.org/rustdoc/ring/hmac/fn.verify.html

src/jar.rs
```diff
@@ -410,95 +419,111 @@ mod secure {
...
     pub fn encrypt_and_sign(key: &[u8], mut cookie: Cookie) -> Cookie {
-        let encrypted_data = encrypt_data(key, &cookie.value);
-        cookie.value = encrypted_data;
-        sign(key, cookie)
-    }
-
-    fn encrypt_data(key: &[u8], val: &str) -> String {
-        let iv = random_iv();
-        let iv_str = iv.to_base64(STANDARD);
+        assert_eq!(key.len(), ENCRYPTION_KEY_LEN);
```
Philipp91: I think _ring_ could benefit from a helper function that handles lines 443-464 in a single call.

frewsxcv: For reference, [here are the lines @Philipp91 mentioned](https://github.com/frewsxcv/cookie-rs/blob/d22fdaf72ebf4606cb1574a57b45f7deef9fb2b4/src/jar.rs#L443-L464).

cc @briansmith in case you find this helpful.

src/jar.rs
frewsxcv: Great idea! Implemented in https://github.com/alexcrichton/cookie-rs/pull/68/commits/cae8d231af68a8543e50db32d243dae0bc539ae2

src/jar.rs
```diff
@@ -88,16 +73,36 @@ pub struct Iter<'a> {
...
-    pub fn new(key: &[u8]) -> CookieJar<'static> {
+    /// The given secret is used to generate keys which are used to sign
+    /// cookies in the signed cookie jar.
+    pub fn new(secret: &[u8]) -> CookieJar<'static> {
+        CookieJar::_new(secret)
+    }
+
+    #[cfg(feature = "secure")]
+    fn _new(secret: &[u8]) -> CookieJar<'static> {
+        let (key256, key512) = secure::generate_keys(secret);
```
SergioBenitez: This `generate_keys` function does a 10k PBKDF2 _each time_ a cookie jar is created. A cookie jar is typically created once per request based on the incoming headers. As such, this is an unacceptable performance regression.

Why do we need to generate keys in this manner? The typical way to do this is to use SHA256/512 to pad the randomness coming in, or demand randomness of a certain length so that it suffices for use as the key. For instance, you could require 512 bits as an input.

src/jar.rs
```diff
@@ -54,31 +53,17 @@ type Read = fn(&Root, Cookie) -> Option<Cookie>;
...
-        // Using a SHA-256 hash to normalize key as Rails suggests.
-        // See https://github.com/rails/rails/blob/master/activesupport/lib/active_support/message_encryptor.rb
-        secure::prepare_key(key)
-    }
-}
-
-#[cfg(not(feature = "secure"))]
-fn prepare_key(_key: &[u8]) -> () {
-    ()
+struct SecureKeys {
```
SergioBenitez: Why do we need two keys of different lengths?

src/jar.rs
```diff
@@ -410,96 +425,105 @@ mod secure {
...
         };
-        cookie.value = text.to_owned();
-
-        let expected = dosign(key, text);
-        if expected.len() != signature.len() ||
-           !memcmp::eq(&expected, &signature) {
-            return None
+        let verification_key =
+            hmac::VerificationKey::new(SIGNING_ALGORITHM, key);
+        let is_valid_signature = hmac::verify(
```
SergioBenitez: It looks like this like and the following few could be chained into something like `.ok().map()`.

src/jar.rs
```diff
@@ -410,96 +425,105 @@ mod secure {
...
+        assert_eq!(key.len(), ENCRYPTION_KEY_LEN);
+        let sealing_key = aead::SealingKey::new(ENCRYPTION_ALGORITHM, key)
+            .expect("could not create aead sealing key");
+        let value_len = cookie.value.as_bytes().len();
+        let overhead_len = ENCRYPTION_ALGORITHM.max_overhead_len();
+
+        // Prepare bytes to be sealed
+        let in_out_len = cookie.value.as_bytes().len() + overhead_len;
+        let mut in_out = vec![0; in_out_len];
+        in_out[..value_len].clone_from_slice(cookie.value.as_bytes());
```
SergioBenitez: Which is this `clone` and not `copy`?

frewsxcv: Accident. Fixed in https://github.com/alexcrichton/cookie-rs/pull/68/commits/e60534daafdd014ac1813164be187273be88ecf2.

src/jar.rs
frewsxcv: It's possible I could do that, but I think the way I wrote it is easier to follow.

SergioBenitez: I disagree, especially because `.ok().map()` is very idiomatic.

frewsxcv: Can you share a concrete example of what you're proposing?

src/jar.rs
frewsxcv: The key size for AEAD ChaCha20 Poly1305 is 256 bits. The key size for HMAC SHA1 is 512 bits.

src/jar.rs
SergioBenitez:
```rust
hmac::verify(&verification_key, text.as_bytes(), &signature).ok()
    .map(|_| {
        cookie.value = text.to_owned();
        cookie
    });
```

The following also works:

```rust
if let Ok(_) = hmac::verify(&verification_key, text.as_bytes(), &signature) {
        cookie.value = text.to_owned();
        Some(cookie)
} else {
    None
} 
```

or even:

```rust
match hmac::verify(&verification_key, text.as_bytes(), &signature) {
    Ok(_) => {
        cookie.value = text.to_owned();
        Some(cookie)
    }
    Err(_) => None
} 
```

frewsxcv: None of those look marginally better to me, but I'll do it anyways.

src/jar.rs
SergioBenitez: Sure, but why not have a 512-bit key and truncate to 256-bits for ChaCha? We're not gaining anything by having two keys here, as far as I can tell; they're both derived from one key.

frewsxcv: I wasn't sure if truncating a 512-bit KDF-derived key would be secure, but people online are telling me it's fine. I'll do that then.

alexcrichton:
Could the crypto stay the same? This is all modeled after Rails is which quite tried and true and I'd prefer to stick to that if possible.

dreid:
@alexcrichton Are you actually concerned about compatibility with existing encrypted cookies? Or do you just want a more familiar set of algorithms?

src/jar.rs
```diff
@@ -410,96 +425,104 @@ mod secure {
...
+        let out_len = try!(
+            aead::open_in_place(&opening_key, &nonce, 0, &mut in_out, &[])
+                .map_err(|_| ()));
+        let decrypted = try!(
+            String::from_utf8(in_out[..out_len].into()).map_err(|_| ()));
+        cookie.value = decrypted;
+        Ok(cookie)
+    }
+
+    pub fn generate_keys(secret: &[u8]) -> ([u8; 32], [u8; 64]) {
```
briansmith: It would be better to return `(aead::SIVSealingKey, hmac::SigningKey)`. Then you could use `aead::seal_in_place()`/`hmac::sign()` and `open_in_place_with_own_key()`/`verify_with_own_key()` for the verification.

src/jar.rs
```diff
@@ -369,32 +374,42 @@ impl<'a> Iterator for Iter<'a> {
...
+    /// Algorithm used to sign the cookie value
+    static SIGNING_ALGORITHM: &'static digest::Algorithm = &digest::SHA1;
+    /// Separator between cookie value and signature
+    static SIGNATURE_SEPARATOR: &'static str = "--";
+    /// Key length (in bytes) used for signing
+    const SIGNING_KEY_LEN: usize = 512 / 8;
+
+    /// Algorithm used to encrypt the cookie value
+    static ENCRYPTION_ALGORITHM: &'static aead::Algorithm =
+        &aead::CHACHA20_POLY1305;
```
briansmith: 1. ChaCha20-Poly1305 isn't an encryption algorithm, it is an AEAD algorithm. In particular, it does encryption AND authentication.

2. ChaCha20-Poly1305 isn't the best choice for this application because 96-bit random nonces aren't a good choice. It would be better to use the upcoming SIV mode AEADs in *ring* and/or XChaCha20 (if we decide to add that). The SIV stuff isn't open source in *ring* yet, but AES-GCM-SIV probably will become open source "soon."

src/jar.rs
```diff
@@ -369,32 +374,42 @@ impl<'a> Iterator for Iter<'a> {
...
-    use self::openssl::{hash, memcmp, symm};
-    use self::openssl::pkey::PKey;
-    use self::openssl::sign::Signer;
-    use self::openssl::hash::MessageDigest;
+    use self::ring::{aead, digest, hmac, rand, pbkdf2};
     use self::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
 
-    pub const MIN_KEY_LEN: usize = 32;
+    /// Algorithm used to sign the cookie value
+    static SIGNING_ALGORITHM: &'static digest::Algorithm = &digest::SHA1;
```
briansmith: 1. SHA-1 Isn't a signing/authentication algorithm. HMAC-SHA1 is.

2. It is not necessary to use HMAC for the authentication-without-encryption case. You can use the AEAD for authentication without encryption by passing `[]` as the plaintext/ciphertext and passing the data to be signed/authenticated in the `ad` parameter. For ChaCha20-Poly1305, this would result in authentication using Poly1305, for example.

sfackler:
@alexcrichton I believe the main motivation here is the switch from OpenSSL to *ring*, which is a less painful dependency to deal with particularly on Windows.

However, *ring* doesn't support AES-CBC, but only AES-GCM and ChaCha20-Poly1305, so we'd need to switch to one of those. One nice property of both of these is that they're AEAD ciphers, meaning that they self-authenticate so we don't have to separately HMAC. They're both well regarded, and are actually the preferred cipher's in Mozilla's Server Side TLS recommendations: https://wiki.mozilla.org/Security/Server_Side_TLS.

@frewsxcv I think AES-256-GCM will be faster than ChaCha20 on hardware with AES-NI. I wonder if it'd make more sense to use that since ~all server hardware will support those?

src/jar.rs
```diff
@@ -410,96 +425,104 @@ mod secure {
...
     }
 
-    fn decrypt_data(key: &[u8], val: &str) -> Option<Vec<u8>> {
-        let (val, iv) = match split_value(val) {
-            Some(pair) => pair, None => return None
-        };
-
-        let actual = match val.from_base64() {
-            Ok(actual) => actual, Err(_) => return None
+    pub fn design_and_decrypt(key: &[u8], mut cookie: Cookie)
```
briansmith: It took me a bit to realize that "design" means "de-sign." The English word for what this does is "authenticate." Also, "open" == "authenticate and decrypt" and "seal" == "encrypt and sign."

src/jar.rs
```diff
@@ -369,32 +374,42 @@ impl<'a> Iterator for Iter<'a> {
...
+    /// Key length (in bytes) used for signing
+    const SIGNING_KEY_LEN: usize = 512 / 8;
+
+    /// Algorithm used to encrypt the cookie value
+    static ENCRYPTION_ALGORITHM: &'static aead::Algorithm =
+        &aead::CHACHA20_POLY1305;
+    /// Separator between sealed cookie value and nonce
+    static SEALED_NONCE_SEPARATOR: &'static str = "--";
+    /// Key length (in bytes) used for encryption
+    const ENCRYPTION_KEY_LEN: usize = 256 / 8;
```
briansmith: You could use `ENCRYPTION_ALGORITHM.key_len()` for this.

src/jar.rs
briansmith: If `secret` is a random value already then HKDF (`ring::hkdf`) should be used to derive the two keys if you insist on still using HMAC for the authentication-only case.

If `secret` is a password then something like PBKDF2 needs to be used to stretch the key material *once*. Then use HKDF to derive the two keys.

alexcrichton:
@dreid oh nah I'm not worried about backcompat here, just the crypto!

@sfackler yeah I'm all for moving to *ring* as the OpenSSL dependency is a bummer. 

My main concern here is that I'm not a cryptographer by any means and I'm hesitant to change from the "accepted practice" of what Rails does which presumably has been tried, true, and vetted for quite awhile. That being said, though, the crypto here is relatively simple IIRC and I'd be open to changing of course! 

Thanks for the review help @sfackler and @briansmith!

mikeycgto:
Hello, I just wanted to pipe in let everyone here know that I wrote a crate called `message_verifier` which providers two main structs, `Verifier` and `Encryptor`, which aim to be compatible with `ActiveSupport::MessageVerifier` and `ActiveSupport::MessageEncryptor`.

My next step was to build a small crate for parsing Rails' session cookies. I plan on writing some blog posts about building Rust micro web services that work with Rails applications. Perhaps `message_verifier` can be integrated with this crate to provide both signed and/or encrypted cookies? By default, Rails' session cookies are both signed and encrypted.

frewsxcv:
Thank you everyone for your feedback ❤️ 

I'm a bit busy for a few days, but I'll try to get around to addressing all the comments this weekend

briansmith:
> My main advice is to wait until we have nonce-misuse-resistant encryption in ring before doing this.

See in particular https://github.com/briansmith/ring/issues/411, https://github.com/briansmith/ring/issues/412, and https://github.com/briansmith/ring/issues/413.

alexcrichton:
Thanks @briansmith!

src/jar.rs
frewsxcv: Regarding the expensive operation, @sfackler had a suggestion to design a `struct Secret` and have it be constructed *once* where it does the PBKDF2 during construction. Then that structure can be passed into `CookieJar` whenever secrets are needed.

frewsxcv:
> > My main advice is to wait until we have nonce-misuse-resistant encryption in ring before doing this.

> See in particular briansmith/ring#411, briansmith/ring#412, and briansmith/ring#413.

I'm still interested in seeing the *ring* migration happen, but considering these comments and my lack of free time right now, I'm not planning on making any changes to this PR for the foreseeable future.

src/jar.rs
briansmith: To be honest, PBKDF2 doesn't make much sense in this application. PBKDF2 is for converting a password into a key. But nobody should be using a password for this stuff. instead, they should be randomly generating a key.

Once you have one key, you can derive multiple keys from it using a KDF such as `ring::hkdf`. This is much faster than PBKDF2 but still not so fast that you want to do it more than necessary. Thus, it is good to create an object for holding the outputs of HKDF.

However, if you go with my suggestion to just use a nonce-misuse-resistant AEAD for both encrypt-and-authenticate and for authenticate-only, then you only need one key, and you can skip using a KDF completely, if you are given a key.

And, again, don't support the mode where one is given a password instead of a key, if you can avoid it.

PR #69, opened 2017-01-18
bump to latest openssl version

davydotcom:


alexcrichton:
Thanks for the PR! I think this isn't necessary though as this is just a semver version requirement, and this essentially means "0.9 compatible, at least 0.9.0". In that sense there's no update necessary! To update locally you can just run `cargo update -p openssl`

PR #70, opened 2017-01-18
Overhaul Cookie Crate

SergioBenitez:
Hello!

This pull request presents the largest change to the crate since its creation. It completely rewrites the `Cookie` structure and API. The major benefits of this rewrite are usability and performance. On the usability side, a structure following the builder pattern has been added as the primary way to construct a `Cookie`. On the performance side, parsing is now completely allocation free, and setting fields in an existing `Cookie` structure can be done in an allocation free way as well.

The summary of changes, from the main commit, is as follows:

  * `Cookie` parsing is allocation-free.
  * All fields in `Cookie` can be set allocation-free.
  * A `Cookie` can have a non-static lifetime.
  * All methods in `Cookie` are documented with examples.
  * A `Cookie` can be built via a new builder structure: `CookieBuilder`.
  * `Cookie` fields are exposed via methods instead of directly.
  * Custom, nonstandard cookie parameters are no longer supported.
  * Additional parsing tests have been added.
  * A suite of formatting tests has been added.
  * A 'percent-encode' feature has been added for automatic percent encoding.
  * Serde and Rustc serialization features have been removed.
  * The top-level crate docs have been improved.
  * The documenation URL now points to `docs.rs`.
  * The entire crate has been `rustfmt`d where possible, removing tabs.
  * The `max_age` field uses a `Duration` instead of an `i64`.

Operationally, I've bumped the version to `0.6.0` to reflect these API changes. I've also added myself as an author, which I felt was fair given that this is a rewrite of a substantial portion of the crate.

Finally, I'd like to request that I be given maintainer access to the repository. I'd really like to continue improving the crate in the future. There are many opportunities for usability and performance improvements in the `CookieJar` API, and I'd like to be able to take the responsibility of seeing them through head-on.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,9 +76,8 @@ pub use parse::ParseError;
 pub use builder::CookieBuilder;
 pub use jar::CookieJar;
 
-#[doc(hidden)]
 #[derive(Debug, Clone)]
-pub enum CookieStr {
+enum CookieStr {
     /// An string derived from indexes (start, end).
     Indexed(usize, usize),
     /// A string derived from a concrete string.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -176,7 +176,6 @@ impl Cookie<'static> {
     /// let cookie = Cookie::new("name", "value");
     /// assert_eq!(cookie.name_value(), ("name", "value"));
     /// ```
-    #[inline(always)]
     pub fn new<N, V>(name: N, value: V) -> Cookie<'static>
         where N: Into<Cow<'static, str>>,
               V: Into<Cow<'static, str>>
@@ -205,7 +204,6 @@ impl Cookie<'static> {
     /// let c = Cookie::build("foo", "bar").finish();
     /// assert_eq!(c.name_value(), ("foo", "bar"));
     /// ```
-    #[inline(always)]
     pub fn build<N, V>(name: N, value: V) -> CookieBuilder
         where N: Into<Cow<'static, str>>,
               V: Into<Cow<'static, str>>
@@ -227,7 +225,6 @@ impl<'c> Cookie<'c> {
     /// assert_eq!(c.name_value(), ("foo", "bar%20baz"));
     /// assert_eq!(c.http_only(), true);
     /// ```
-    #[inline]
     pub fn parse<S>(s: S) -> Result<Cookie<'c>, ParseError>
         where S: Into<Cow<'c, str>>
     {
@@ -247,7 +244,6 @@ impl<'c> Cookie<'c> {
     /// assert_eq!(c.name_value(), ("foo", "bar baz"));
     /// assert_eq!(c.http_only(), true);
     /// ```
-    #[inline]
     #[cfg(feature = "percent-encode")]
     pub fn parse_encoded<S>(s: S) -> Result<Cookie<'c>, ParseError>
         where S: Into<Cow<'c, str>>
@@ -345,7 +341,7 @@ impl<'c> Cookie<'c> {
     /// let c = Cookie::new("name", "value");
     /// assert_eq!(c.name_value(), ("name", "value"));
     /// ```
-    #[inline(always)]
+    #[inline]
     pub fn name_value(&self) -> (&str, &str) {
         (self.name(), self.value())
     }
@@ -476,7 +472,6 @@ impl<'c> Cookie<'c> {
     /// c.set_name("foo");
     /// assert_eq!(c.name(), "foo");
     /// ```
-    #[inline(always)]
     pub fn set_name<N: Into<Cow<'static, str>>>(&mut self, name: N) {
         self.name = CookieStr::Concrete(name.into())
     }
@@ -494,7 +489,6 @@ impl<'c> Cookie<'c> {
     /// c.set_value("bar");
     /// assert_eq!(c.value(), "bar");
     /// ```
-    #[inline(always)]
     pub fn set_value<V: Into<Cow<'static, str>>>(&mut self, value: V) {
         self.value = CookieStr::Concrete(value.into())
     }
@@ -512,7 +506,7 @@ impl<'c> Cookie<'c> {
     /// c.set_http_only(true);
     /// assert_eq!(c.http_only(), true);
     /// ```
-    #[inline(always)]
+    #[inline]
     pub fn set_http_only(&mut self, value: bool) {
         self.http_only = value;
     }
@@ -530,7 +524,7 @@ impl<'c> Cookie<'c> {
     /// c.set_secure(true);
     /// assert_eq!(c.secure(), true);
     /// ```
-    #[inline(always)]
+    #[inline]
     pub fn set_secure(&mut self, value: bool) {
         self.secure = value;
     }
@@ -554,7 +548,7 @@ impl<'c> Cookie<'c> {
     /// assert_eq!(c.max_age(), Some(Duration::hours(10)));
     /// # }
     /// ```
-    #[inline(always)]
+    #[inline]
     pub fn set_max_age(&mut self, value: Duration) {
         self.max_age = Some(value);
     }
@@ -572,7 +566,6 @@ impl<'c> Cookie<'c> {
     /// c.set_path("/");
     /// assert_eq!(c.path(), Some("/"));
     /// ```
-    #[inline(always)]
     pub fn set_path<P: Into<Cow<'static, str>>>(&mut self, path: P) {
         self.path = Some(CookieStr::Concrete(path.into()));
     }
@@ -590,7 +583,6 @@ impl<'c> Cookie<'c> {
     /// c.set_domain("rust-lang.org");
     /// assert_eq!(c.domain(), Some("rust-lang.org"));
     /// ```
-    #[inline(always)]
     pub fn set_domain<D: Into<Cow<'static, str>>>(&mut self, domain: D) {
         self.domain = Some(CookieStr::Concrete(domain.into()));
     }
@@ -616,7 +608,7 @@ impl<'c> Cookie<'c> {
     /// assert!(c.expires().is_some())
     /// # }
     /// ```
-    #[inline(always)]
+    #[inline]
     pub fn set_expires(&mut self, time: Tm) {
         self.expires = Some(time);
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,16 +1,19 @@
-#![deny(missing_docs)]
-#![doc(html_root_url = "http://alexcrichton.com/cookie-rs")]
-#![cfg_attr(test, deny(warnings))]
+// #![deny(missing_docs)]
+// #![cfg_attr(test, deny(warnings))]
 
-//! HTTP Cookie parsing and Cookie Jar management
+//! HTTP Cookie parsing and Cookie Jar management.
 //!
-//! Usage:
+//! This crates provides the [Cookie](struct.Cookie.html) type, which directly
+//! maps to an HTTP cookie, and the [CookieJar](struct.CookieJar.html) type,
+//! which allows for simple management of many cookies as well as encryption and
+//! signing of cookies for session management.
 //!
-//! add the following to the `[dependencies]` section of
-//! your `Cargo.toml`:
+//! # Usage
+//!
+//! Add the following to the `[dependencies]` section of your `Cargo.toml`:
 //!
 //! ```ignore
-//! cookie = "0.4"
+//! cookie = "0.6"
 //! ```
 //!
 //! Then add the following line to your crate root:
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
+//!
+//! * **secure** (enabled by default)
+//!
+//!   Enables signing and encryption of cookies.
+//!
+//!   When this feature is enabled, signed and encrypted cookies jars will
+//!   encrypt and/or sign any cookies added to them. When this feature is
+//!   disabled, those cookies will be added in plaintext.
+//!
+//! * **percent-encode** (disabled by default)
```
alexcrichton: I like that this is a feature to enable in the crate as some applications will need it, but I think it may not be appropriate as a Cargo (compile-time feature). If the `cookie` library is a shared dependency between two crates where one was percent encoding and the other doesn't they'll both get it (and the crate that doesn't want it may not operate well).

Perhaps this could be a configuration option during parsing?

src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
-    pub path: Option<String>,
-    #[allow(missing_docs)]
-    pub secure: bool,
-    #[allow(missing_docs)]
-    pub httponly: bool,
-    #[allow(missing_docs)]
-    pub custom: BTreeMap<String, String>,
+#[doc(hidden)]
+#[derive(Debug, Clone)]
+pub enum CookieStr {
```
alexcrichton: Would it be possible to hide this type? It seems like it'd be nice to avoid exposing if possible

src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
-    /// let c = Cookie::new("foo".into(), "bar".into());
-    /// assert_eq!(c.name, "foo");
-    /// assert_eq!(c.value, "bar");
+    /// let cookie = Cookie::new("name", "value");
+    /// assert_eq!(cookie.name_value(), ("name", "value"));
     /// ```
-    pub fn new(name: String, value: String) -> Cookie {
+    #[inline(always)]
+    pub fn new<N, V>(name: N, value: V) -> Cookie<'static>
+        where N: Into<Cow<'static, str>>,
```
alexcrichton: I've seen the `Into<Cow<'static, str>>` bounds here a lot, but I fear for the ergonomics of this. Specifically this fails to compile:

```rust
fn foo<T: Into<Cow<'static, str>>>(_t: T) {}
                                            
fn main() {                                 
    let a = String::new();                  
    foo(&a[..]);                            
}                                           
```

Namely this can't be called with `&str` where it's not `'static`. that means that this can only be called with the `String` type or with string literals. In terms of ergonomics I wonder if it'd be nice to change this to `Into<String>` (or just `&str`) perhaps?

Is there much of a performance gain to accepting `&'static str` without allocating? I would imagine that the biggest perf win here is indexed strings keeping track of the original cookie rather than allocating separate substrings. For creating/configuring fields I'd imagine that there's not much perf to be regained in real-world use cases perhaps?

src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
+///
+/// let cookie = Cookie::build("name", "value")
+///     .domain("www.rust-lang.org")
+///     .path("/")
+///     .secure(true)
+///     .http_only(true)
+///     .finish();
+/// ```
+#[derive(Debug, Clone)]
+pub struct Cookie<'c> {
```
alexcrichton: Could you elaborate on the case for a non-static lifetime here? I'd expect this to be `'static` almost all the time as threading a borrow through all the way to the end could be quite difficult.

I wonder if we could expose a separate parser type which explicitly works over borrowed values?

src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let c = Cookie::build("foo", "bar").finish();
+    /// assert_eq!(c.name_value(), ("foo", "bar"));
+    /// ```
+    #[inline(always)]
+    pub fn build<N, V>(name: N, value: V) -> CookieBuilder
```
alexcrichton: I'm personally not the biggest fan of this pattern (I know others like, though), but could this be eschewed in favor of `CookieBuilder::new`?

src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let c = Cookie::build("foo", "bar").finish();
+    /// assert_eq!(c.name_value(), ("foo", "bar"));
+    /// ```
+    #[inline(always)]
```
alexcrichton: `#[inline(always)]` should basically always be omitted until proven otherwise through profiling, could these annotations be omitted?

src/lib.rs
```diff
@@ -1,427 +1,749 @@
...
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let c = Cookie::parse("foo=bar; HttpOnly").unwrap();
+    /// assert_eq!(c.name_value(), ("foo", "bar"));
+    /// assert_eq!(c.http_only(), true);
+    /// ```
+    #[inline]
```
alexcrichton: Do functions like this really need to be inlined?

src/parse.rs
```diff
@@ -0,0 +1,335 @@
...
+pub enum ParseError {
+    /// The cookie did not contain a name/value pair.
+    MissingPair,
+    /// The cookie's name was empty.
+    EmptyName,
+    /// Decoding the cookie's name or value resulted in invalid UTF-8.
+    Utf8Error(Utf8Error),
+    /// Internal error: this should not occur.
+    Internal,
+}
```
alexcrichton: Could you throw a `__Nonexhaustive` case in here to discourage exhaustive matching?

src/lib.rs
SergioBenitez: Is that really what Cargo would do in that scenario? That's really surprising behavior from Cargo in that case. I would expect that if I include `cookie` as a direct dependency in my project, then it's configured at build-time exactly the way I specified and doesn't inherit features specified from other dependencies. Are you sure that's what happens? Or maybe you're referring to a different scenario?

src/lib.rs
SergioBenitez: Turns out yes!

src/lib.rs
SergioBenitez: So the idea is that you might want to parse a non-static `str` as a `Cookie` without any allocations. Without the lifetime, this simply isn't possible.

You don't need to thread the borrow at all; this is what the `into_owned` method was written to accomplish. It allow you to extend the lifetime of the `Cookie` to `static` with at-most one allocation.

src/lib.rs
SergioBenitez: In your example, the solution is straightforward: either `clone` the `String`, or move the `String` into `foo`: `foo(a.clone())` or `foo(a)`. I'd expect any Rust programmer above an introductory level to see this. And if not, it's something that's easily learned or remedied with a few words in the documentation. As a point of evidence, Rocket uses these bounds extensively without a single complaint from users, even though many of Rocket's users are first-time Rustaceans.

The current API gives the user a choice: an allocation is only required when necessary. Changing the bounds to `Into<String>` means that setting any string-typed field results in one allocation.

Regarding performance, the `Into<Cow<'static, str>>` bounds will always be at least as performant as using `Into<String>` as it doesn't _require_ allocations. It's hard to argue whether users will pass `&'static str` frequently enough to justify this, but Rocket will certainly be using `&'static str` values almost exclusively, and I'd expect at least cookie names to be `&'static str`s.

src/lib.rs
SergioBenitez: All of them?

I feel that these are evidently okay. I only use `#[inline(always)]` when the optimizer _should_ be inlining the function call anyway since the call would be much more expensive that simply executing the code inline. For instance, if setting or retrieving a field is _just_ reading or writing a field, then that's the code that should be emitted.

I've benchmarked whether this actually matters in other libraries, and the answer is yes; it matters quite a bit. I've seen anywhere from 20 to 40% improvements by using `#[inline(always)]`.

src/lib.rs
SergioBenitez: What's wrong with keeping the method here? If you don't like it, you don't have to use it. But I think asking users to import `CookieBuilder` when it's wholly unnecessary isn't great.

src/lib.rs
SergioBenitez: This should actually be `#[inline(always)]`, but I missed it.

I'm not sure what you mean, though. Are you saying that you'd expect the optimizer to inline it anyway without the annotation? That's what I'd expect, too, but I've seen it not do it.

src/parse.rs
SergioBenitez: Why do we want to discourage exhaustive matching? I don't mean to argue, I'm just genuinely curious.

src/lib.rs
seanmonstar: No, cargo features are additive. On one crate, you say "I want cookie and only need this functionality", and another can say "I want cookie, but only need that other part". Cargo will compile only 1 version, with all the request features enabled.

They are not meant to be used to **change** behavior.

src/lib.rs
seanmonstar: In general, it's preferred to have, if anything, simple `#[inline]`. This adds the function as a possible inlineable function across crate bounds, but still allows the optimizer's heuristics to determine if it would be an improvement, versus a possible explosion in code side or whatever.

src/parse.rs
seanmonstar: It means that if you want to add an additional error case, it is automatically a breaking change, since it would break all exhaustive matches.

```rust
match e {
    MissingPair => (),
    EmptyName => (),
    Utf8Error(..) => (),
    Internal => (),
}
```

The above would break if we added a new variant, `SomeOtherEdgeCaseWeLearnedAbout`. The `__Nonexhaustive` (or `__DontMatchMe`) pattern makes it obvious to any writer that they are taking the breakage into their own hands:

```rust
match e {
    MissingPair => (),
    EmptyName => (),
    Utf8Error(..) => (),
    Internal => (),
    __Nonexhasutive => (), // shame on me
    // _ => () // <- the better way
}
```

src/lib.rs
alexcrichton: Heh as the author of Cargo, I can safely say this is what happens.

Cargo only compiles crates once, and any number of other crates can be depending on that crate. All activate features are unioned together and the crate is compiled once with all those features.

src/parse.rs
alexcrichton: @SergioBenitez yeah as @seanmonstar said it means that we can add more error cases in the future without it being a breaking change.

src/lib.rs
alexcrichton: There's no convention for this in the standard library, for example. This is a free function on a type returning a completely different type. Although it's a slight win ergonomic style I don't think it's worth the confusion. I'm personally always confused when I see `Foo::bar()` and it doesn't return a `Foo` in one form or another.

src/lib.rs
alexcrichton: @SergioBenitez yes I believe all of them should be removed. When you mention 20-40% improvements is that for this library specifically? Or is that for other libraries? Conclusions like that tend to not be true across libraries/situations, I've found.

Are you familiar with the inlining heuristics in the compiler today? Or in other words what this is actually doing? (that helps in understanding why it's not necessary most of the time)

src/lib.rs
alexcrichton: Sorry I think my point may have been lost, let me rephrase:

* This API does not accept `&String`
* This API does not accept `&str` where it's not a string literal
* This API only protects against allocations where otherwise you have a string literal

With `Into<Cow>` the ergonomics aren't as good as just taking `&str` but it has the upside of not requiring an allocation with string literals. Do you think it's worth the tradeoff?

src/lib.rs
alexcrichton: What situations are you thinking of where you want to parse a cookie and *not* call `into_owned`?

src/lib.rs
seanmonstar: I understand the dislike of this pattern, I personally favor it slightly more than requiring another import. I feel there is precedence in the standard library with the format builders. `formatter.debug_struct("Foo") -> DebugStructBuilder`, instead of `DebugStructBuilder::new(formatter, "Foo")`.

src/lib.rs
SergioBenitez: Sorry, I didn't mean to imply that you don't know what's going on, but rather that I may be referring to a different scenario than you were.

This is, frankly, really surprising behavior, and it's the first I'm learning of it. None of the material on features makes it clear that this they are additive and can cause conflicts if not used purely in this manner. The Cargo documentation refers to features as "conditional compilation" only.

In any case, with this new knowledge, I completely agree that this shouldn't be a feature. But this also can't be just another parameter during parsing as it also affects the `Display` of a cookie. Which begs the question: should the `Display` implementation encode the name/value or not? I think either choice is arbitrary. Maybe there shouldn't be a `Display` implementation?

src/lib.rs
SergioBenitez: If you're parsing a cookie from an allocated `String`, you don't need to call `into_owned`; you simply use `Cookie::parse_static(String)`.

The situation I have in mind is one where headers are stored in some structure and a borrow is made to the value of the `Cookie` header. You shouldn't need to reallocate unnecessarily in this case.

src/lib.rs
SergioBenitez: I don't think your point has been lost at all. I'm simply saying that these things aren't things to be concerned about since you can always call `.to_string()` or `clone()`. Further, your third point is incorrect; this API also protects against an additional allocation when you already have a `String` that you can move, something taking an `&str` wouldn't.

Personally, I would like some magical `Into<StaticString>` type that would let me take any `&str` and `String` and allocate only when an `&str` with a non-static lifetime was passed in. Unfortunately, this is currently impossible. As such, I think the tradeoff, which I believe is minor, is entirely worth it.

src/lib.rs
SergioBenitez: I'm not familiar with heuristics inside of `rustc`, no. But I'm aware of what this is actually _trying_ to do. My understanding is that it does just as it implies: it asks the compile to _always_ inline a call to the function. This is why I applied the annotation.

I was referring to improvements in other libraries, not this one. I benchmarked building a `Cookie` with and without the `#[inline(always)]`, and there is no difference, so it seems prudent to remove the annotation. Should I still use `#[inline]` to encourage inlining across crates?

src/lib.rs
SergioBenitez: @alexcrichton I've seen this pattern applied broadly across the ecosystem. `build` _is_ returning a `Cookie` in some form, though; a `Builder` version of it. Rocket uses this convention broadly, if it makes any difference, with only positive feedback.

src/lib.rs
alexcrichton: I'd probably stick with avoiding interpretation of percent encoding to represent the "true value" underneath the hood perhaps?

src/lib.rs
alexcrichton: Hm ok, that makes sense. It'd be nice if we could cut down on the extra methods here in favor of maybe just one concept. I wonder if we could leverage the `TryFrom` trait and such? That seems like it'd be able to carry through lifetimes and such and then could just implement:

```rust
impl<'a> TryFrom<Cow<'a, str>> for Cookie<'a> { ... }
impl<'a> TryFrom<&'a str> for Cookie<'a> { ... }
impl<'a> TryFrom<String> for Cookie<'a> { ... }
```

etc

I'm a little worried about exposing too many knobs as they can end up being confusing with so many entry points.

src/lib.rs
alexcrichton: In general you never need `#[inline]` as LLVM's quite good at determining what needs to be inlined and what doesn't be. If LLVM doesn't inline a small function it tends to have a very good reason not do, so forcing it to go against that decision with an annotation like `#[inline(always)]` typically needs to be done with great care. We've had pathological cases in the past where an entire library was plastered with `#[inline(always)]` and it took ages to compile and generated huge binaries (both fixed by removing the annotations).

Now there's one case (as you've mentioned) that LLVM cannot inline, which is when you're translating across crates. Fundamentally LLVM does not have the code on hand (e.g. it's not in the translation unit) so it wouldn't know how to inline. This is one primary reason why `#[inline]` exists, to say that an otherwise non-inlineable function should be inlined across crates. This causes the compiler to encode the MIR for the function in metadata, and whenever used the compiler will re-codegen the function. This can have large impacts on compile time if used too liberally.

So if the purpose here is simply to inline across crates, the question is then if the attribute is needed to do that. Most of these functions, however, are generic, which means that the attribute isn't necessary. Generic functions are already inlined in all downstream crates when called (e.g. encoded in MIR and retranslated when instantiated). 

All in all a good rule of thumb is:

* Generic functions never need `#[inline]` tags. Unless you profile and see the compiler *not* inlining, you should always omit them.
* Concrete (non-generic functions) only need tags to inline across crates. My recommendation would also be to witness a benchmark that actually speeds up with the inlining.

---

Another historical thing that `#[inline]` does is that it also "hints" to LLVM that the function should be inlined. This means it does two operations: inline across crates and provide LLVM a strong recommendation it should be inlined. We never got around to adding a "only inline this in other crates" annotation, but it means that usage of `#[inline]` on generic functions can hurt downstream compile times by forcing too much to be inlined that wouldn't otherwise be inlined.

The use case for `#[inline(always)]` is actually quite rare from what I've seen. I've literally never seen a case of LLVM not inlining a function when I thought it should do so (when it could, of course). This sort of annotation is typically useful for very unsafe code dealing with stack frames and you've got to guarantee that a stack frame for the function doesn't exist. This sometimes is useful in contexts like kernels or stack trace programs (for example).

src/lib.rs
alexcrichton: I wonder if perhaps we could go for an even more ergonomic API perhaps? The builder already contains a valid `Cookie` and basically just renames APIs, so maybe:

```rust
let cookie = Cookie::new(a, b).builder().foo(..).bar(..).build();
```

That'd also allow you to use a builder style to set fields at any time?

Alternatively we could just have the `set` methods return `&mut Self` and build that way.

src/parse.rs
```diff
@@ -0,0 +1,340 @@
...
+        _ => return Err(ParseError::Internal),
+    };
+
+    if name.is_empty() {
+        return Err(ParseError::EmptyName);
+    }
+
+    // Create a cookie with all of the defaults. We'll fill things in while we
+    // iterate through the parameters below.
+    let (name, value) = name_val(name, value, s)?;
```
seanmonstar: As much fun as it is to use `?`, that does mean that the minimum version compiler of rustc is increased with this change. I'm all for people using it in their applications, but I've personally so far stuck with `try!`, which allows the libraries to be used with the most amount of compilers. I'd personally rather if the cookie crate did the same.

src/lib.rs
SergioBenitez: I'm not sure I follow what you're proposing. Are you saying `cookie` should simply not percent-encode/decode at all? The issue with this decision is that percent-encoding/decoding only the name/value pairs outside of the cookie crate is _really_ cumbersome on the encode side. It essentially requires you to reimplement `Display` or clone the cookie (+ allocations) and reset the fields. Neither are great solutions, unfortunately.

src/lib.rs
SergioBenitez: `TryFrom` is still unstable, unfortunately. But even then, we wouldn't be able to get the same level of flexibility with a single thing like this because of the lifetime on ``a`. We'd need to have a case of `impl TryFrom<&'static str>` _and_ `impl<'a> TryForm<&'a str>`, but this isn't allowed, of course.

src/lib.rs
SergioBenitez: Awesome! Thanks for sharing this.

It sounds like your recommendation, then, is to remove all `inline`s from generic functions, and use only `#[inline]` where appropriate for non-generic functions. Is that right?

src/lib.rs
SergioBenitez: We definitely want to be able to own the `Cookie` value after building it, since it'll be passed into a `CookieJar` by move. Going the `&mut self` route would prevent this.

I'm against the `.builder()` method for a couple of reasons. First, I haven't seen this convention used before, but I have seen the `T::build()` convention used quite a bit. Second, it adds an extra invocation when building for no reason. It also means we can't have getters with the name of the field. IE, we can't have both `foo(T)` and `foo()`.

I _really_ like the `build()` approach to this.

src/parse.rs
SergioBenitez: I don't think we have a legacy issue in Rust, seeing as releases are rolling and always backwards compatible. We're approaching `1.15`, which means the `?` syntax has been stable for two releases, so I'd be very surprised if this was an issue.

I really prefer using `?`, but I don't have an opinion otherwise.

seanmonstar: I prefer using it as well! I've also grown sympathetic to those in situations where auditing a tool upgrade is a longer process than a simple rustup. In this case, being just a style difference, it seems on the simpler side to me to stay usable to people who for whatever reason cannot upgrade their compiler yet.

src/lib.rs
alexcrichton: Oh no I'm fine supporting percent encoding and all, we'll just have to have separate methods that say "yes I'd like to decode/encode this". Implementations like `Display` have to pick one, and I think avoiding interpretation of the value is the best choice because then it means you can just throw a `.to_string()` and ship it to a client.

src/lib.rs
alexcrichton: My worry is that so many entry points, while motivated, can be confusing. It sounds like though that they should all stay to start off with and we can figure out ergonomics later.

src/lib.rs
alexcrichton: Yes `inline` basically isn't necessary on generics, and I wouldn't recommend placing `#[inline]` on non-generic functions until proven otherwise. Rust already has compile-time issues and liberal application of `#[inline]` tends to exascerbate it even more.

src/lib.rs
alexcrichton: Ok let's just stick with that for now then.

src/parse.rs
alexcrichton: My rule of thumb is:

* Always work on the current stable
* Try to always work on the previous current stable
* Support older versions only if requested.

In that sense using `?` I think fits the bill, so if we get future PRs to move the version requirement back we can deal with it then.

src/lib.rs
SergioBenitez: Okay, I've addresses this by adding two new methods: `encoded` and `parse_encoded`.

src/lib.rs
SergioBenitez: I've removed `parse_static` in favor of a generic `parse` than should result in a `static` lifetime when possible/coerced.

SergioBenitez:
I believe the last two commits resolves any outstanding comments. Looking forward to getting this merged!

alexcrichton:
Looks good to me, thanks!

PR #71, opened 2017-01-26
Fix parser.

twittner:
According to RFC 6265 cookie values may contain '=' characters. The
current parser implementation only considers a cookie value up to the
first occurrence of '=' and ignores the rest.

This commit splits the parse string only at the first position of '='
and retains the cookie value as is.


alexcrichton:
Thanks! Could you be sure to add a test for this as well?

twittner:
The PR has been updated to include a simple test case.

SergioBenitez:
@alexcrichton Did you see the test that was added? The doesn't actually check the functionality that was added.

PR #76, opened 2017-02-15
Overhaul `CookieJar`: more efficient, more secure, more docs.

### LINKED ISSUE #35: A CookieJar cannot be shared between threads, even with a Mutex
``` rust
use std::sync::{Arc, Mutex};
use std::thread;

extern crate cookie;
use cookie::{Cookie, CookieJar};

fn main() {
    let m = Arc::new(Mutex::new(CookieJar::new(b"f8f9eaf1ecdedff5e5b749c58115441e")));

    thread::spawn(move || {
      m
    });
}
```

This fails to compile with the following error message:

```
src/main.rs:10:5: 10:18 error: the trait `core::marker::Sync` is not implemented for the type `core::cell::UnsafeCell<std::collections::hash::map::HashMap<collections::string::String, cookie::Cookie>>` [E0277]
src/main.rs:10     thread::spawn(move || {
                   ^~~~~~~~~~~~~
src/main.rs:10:5: 10:18 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:10:5: 10:18 note: `core::cell::UnsafeCell<std::collections::hash::map::HashMap<collections::string::String, cookie::Cookie>>` cannot be shared between threads safely
src/main.rs:10     thread::spawn(move || {
                   ^~~~~~~~~~~~~
src/main.rs:10:5: 10:18 error: the trait `core::marker::Sync` is not implemented for the type `core::cell::UnsafeCell<std::collections::hash::set::HashSet<collections::string::String>>` [E0277]
src/main.rs:10     thread::spawn(move || {
                   ^~~~~~~~~~~~~
src/main.rs:10:5: 10:18 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:10:5: 10:18 note: `core::cell::UnsafeCell<std::collections::hash::set::HashSet<collections::string::String>>` cannot be shared between threads safely
src/main.rs:10     thread::spawn(move || {
                   ^~~~~~~~~~~~~
src/main.rs:10:5: 10:18 error: the trait `core::marker::Sync` is not implemented for the type `core::cell::UnsafeCell<usize>` [E0277]
src/main.rs:10     thread::spawn(move || {
                   ^~~~~~~~~~~~~
src/main.rs:10:5: 10:18 help: run `rustc --explain E0277` to see a detailed explanation
src/main.rs:10:5: 10:18 note: `core::cell::UnsafeCell<usize>` cannot be shared between threads safely
src/main.rs:10     thread::spawn(move || {
```

I've been looking for a consumer-side solution for over an hour now, but couldn't find one. Is the problem with `CookieJar` itself?

### LINKED ISSUE #50: Use an AEAD cipher instead of AES-CBC + HMAC-SHA1

### LINKED ISSUE #64: Implement Send?
Can/should the `CookieJar` type (unsafely) implement `Send`? Or perhaps one of its component types. It's rather prohibitive right now that it doesn't.

### LINKED ISSUE #75: Remove cookies are missing Path and Domain
When a cookie is removed from the Jar, it only records the cookie's name. So when the Set-Cookie header is generated to clear the cookie, the header does not have the Path or Domain parameters, even if the original cookie had them. This is a problem because the new Set-Cookie header only overwrites the previous cookie if the Path and Domain parameters match.

SergioBenitez:
This commit completely rewrites `CookieJar` so that it is more efficient while being simpler and more correct overall. At a high-level, a `CookieJar` now returns borrows to `Cookie` structures whenever possible. `CookieJar` and children jars (now `PrivateJar` and `SignedJar`) are distinct types. The "encrypted" child jar has been replaced by `PrivateJar` which uses authenticated encryption. All crypto is done via `ring`.

A detailed list of changes is below:

  * The "permanent" child jar has been removed.
  * A `make_permanent` method has been added to `Cookie`.
  * A `permanent` method has been added to `CookieBuilder`.
  * A `named` constructor has been added to `Cookie`.
  * `CookieJar` no longer has a generic lifetime parameter.
  * A `CookieJar` no longer contains `RefCell`s.
  * `CookieJar::{add_original, add, remove, clear}` take `&mut self`.
  * `CookieJar::new` does not take in a `key` parameter.
  * `CookieJar::find` returns a borrow to a `Cookie`.
  * `CookieJar::remove` takes in a `Cookie`.
  * `CookieJar::clear` has been deprecated.
  * A `CookieJar` tracks its modifications more robustly, passing a new suite of unit tests.
  * Computing the `delta` is zero-cost operation.
  * Cookie iteration is significantly simplified.
  * The `signed` method has been removed from `CookieJar`.
  * A new `Signed` trait was added with a `signed(&key)` method.
  * The `Signed` trait is implemented for `CookieJar`.
  * `SignedJar` cookies sign via ring's HMAC-SHA256 implementation.
  * The `encrypted` method has been removed from `CookieJar`.
  * A new `Private` trait was added with a `private(&key)` method.
  * The `Private` trait is implemented for `CookieJar`.
  * `PrivateJar` cookies use AEAD (AES-256-GCM) from ring.
  * The `openssl` dependency was dropped.
  * A `ring` dependency was introduced.
  * All items are extensively documented.

Fixes #35.
Fixes #50.
Fixes #64.
Closes #68.
Fixes #75.

Finally, I'd like to, once again, request that I be given maintainer access to the repository/crate. I'd really like to continue improving and maintaining this crate in the future.

### DIFF
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -3,6 +3,8 @@ use std::mem::replace;
 
 use time::{self, Duration};
 
+#[cfg(feature = "secure")]
+use secure::{PrivateJar, SignedJar};
 use delta::DeltaCookie;
 use Cookie;
 
@@ -305,6 +307,93 @@ impl CookieJar {
     pub fn iter(&self) -> Iter {
         Iter { delta_cookies: self.delta_cookies.union(&self.original_cookies) }
     }
+
+    /// Returns a `PrivateJar` with `self` as its parent jar using the key `key`
+    /// to sign/encrypt and verify/decrypt cookies added/retrieved from the
+    /// child jar. The key must be exactly 32 bytes. For security, the key
+    /// _must_ be cryptographically random.
+    ///
+    /// Any modifications to the child jar will be reflected on the parent jar,
+    /// and any retrievals from the child jar will be made from the parent jar.
+    ///
+    /// This method is only available when the `secure` feature is enabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `key` is not exactly 32 bytes long.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use cookie::{Cookie, CookieJar};
+    ///
+    /// // We use a bogus key for demonstration purposes.
+    /// let key: Vec<_> = (0..32).collect();
+    ///
+    /// // Add a private (signed + encrypted) cookie.
+    /// let mut jar = CookieJar::new();
+    /// jar.private(&key).add(Cookie::new("private", "text"));
+    ///
+    /// // The cookie's contents are encrypted.
+    /// assert_ne!(jar.get("private").unwrap().value(), "text");
+    ///
+    /// // They can be decrypted and verified through the child jar.
+    /// assert_eq!(jar.private(&key).get("private").unwrap().value(), "text");
+    ///
+    /// // A tampered with cookie does not validate but still exists.
+    /// let mut cookie = jar.get("private").unwrap().clone();
+    /// jar.add(Cookie::new("private", cookie.value().to_string() + "!"));
+    /// assert!(jar.private(&key).get("private").is_none());
+    /// assert!(jar.get("private").is_some());
+    /// ```
+    #[cfg(feature = "secure")]
+    pub fn private<'a, 'k>(&'a mut self, key: &'k [u8]) -> PrivateJar<'a, 'k> {
+        PrivateJar::new(self, key)
+    }
+
+    /// Returns a `SignedJar` with `self` as its parent jar using the key `key`
+    /// to sign/verify cookies added/retrieved from the child jar. The key must
+    /// be exactly 64 bytes. For security, the key _must_ be cryptographically
+    /// random.
+    ///
+    /// Any modifications to the child jar will be reflected on the parent jar,
+    /// and any retrievals from the child jar will be made from the parent jar.
+    ///
+    /// This method is only available when the `secure` feature is enabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `key` is not exactly 64 bytes long.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use cookie::{Cookie, CookieJar};
+    ///
+    /// // We use a bogus key for demonstration purposes.
+    /// let key: Vec<_> = (0..64).collect();
+    ///
+    /// // Add a signed cookie.
+    /// let mut jar = CookieJar::new();
+    /// jar.signed(&key).add(Cookie::new("signed", "text"));
+    ///
+    /// // The cookie's contents are signed but still in plaintext.
+    /// assert_ne!(jar.get("signed").unwrap().value(), "text");
+    /// assert!(jar.get("signed").unwrap().value().contains("text"));
+    ///
+    /// // They can be verified through the child jar.
+    /// assert_eq!(jar.signed(&key).get("signed").unwrap().value(), "text");
+    ///
+    /// // A tampered with cookie does not validate but still exists.
+    /// let mut cookie = jar.get("signed").unwrap().clone();
+    /// jar.add(Cookie::new("signed", cookie.value().to_string() + "!"));
+    /// assert!(jar.signed(&key).get("signed").is_none());
+    /// assert!(jar.get("signed").is_some());
+    /// ```
+    #[cfg(feature = "secure")]
+    pub fn signed<'a, 'k>(&'a mut self, key: &'k [u8]) -> SignedJar<'a, 'k> {
+        SignedJar::new(self, key)
+    }
 }
 
 use std::collections::hash_set::Iter as HashSetIter;
@@ -348,9 +437,6 @@ impl<'a> Iterator for Iter<'a> {
 mod test {
     use super::CookieJar;
     use Cookie;
-
-    #[cfg(feature = "secure")]
-    use secure::{Signed, Private};
 
     #[test]
     #[allow(deprecated)]
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -1,540 +1,365 @@
-//! A cookie jar implementation for storing a set of cookies.
-//!
-//! This CookieJar type can be used to manage a session of cookies by keeping
-//! track of cookies that are added and deleted over time. It provides a method,
-//! `delta`, which will calculate the number of `Set-Cookie` headers that need
-//! to be sent back to a client which tracks the changes in the lifetime of the
-//! jar itself.
-//!
-//! A cookie jar can also be borrowed to a child cookie jar with new
-//! functionality such as automatically signing cookies, storing permanent
-//! cookies, etc. This functionality can also be chained together.
-
-use std::collections::{HashMap, HashSet};
-use std::cell::RefCell;
-use std::fmt;
-use std::borrow::Cow;
+use std::collections::HashSet;
+use std::mem::replace;
 
 use time::{self, Duration};
 
-use ::Cookie;
+use delta::DeltaCookie;
+use Cookie;
 
-/// A jar of cookies for managing a session.
+/// A collection of cookies that tracks its modifications.
 ///
-/// # Example
+/// A `CookieJar` provides storage for any number of cookies. Any changes made
+/// to the jar are tracked; the changes can be retrieved via the
+/// [delta](#method.delta) method which returns an interator over the changes.
 ///
-/// ```
+/// # Usage
+///
+/// A jar's life begins via [new](#method.new) and calls to
+/// [add_original](#method.add_original):
+///
+/// ```rust
 /// use cookie::{Cookie, CookieJar};
 ///
-/// let c = CookieJar::new(b"f8f9eaf1ecdedff5e5b749c58115441e");
-///
-/// // Add a cookie to this jar
-/// c.add(Cookie::new("key".to_string(), "value".to_string()));
-///
-/// // Remove the added cookie
-/// c.remove("key");
+/// let mut jar = CookieJar::new();
+/// jar.add_original(Cookie::new("name", "value"));
+/// jar.add_original(Cookie::new("second", "another"));
 /// ```
-pub struct CookieJar<'a> {
-    flavor: Flavor<'a>,
+///
+/// Cookies can be added via [add](#method.add) and removed via
+/// [remove](#method.remove). Finally, cookies can be looked up via
+/// [find](#method.find):
+///
+/// ```rust
+/// # use cookie::{Cookie, CookieJar};
+/// let mut jar = CookieJar::new();
+/// jar.add(Cookie::new("a", "one"));
+/// jar.add(Cookie::new("b", "two"));
+///
+/// assert_eq!(jar.find("a").map(|c| c.value()), Some("one"));
+/// assert_eq!(jar.find("b").map(|c| c.value()), Some("two"));
+///
+/// jar.remove(Cookie::named("b"));
+/// assert!(jar.find("b").is_none());
+/// ```
+///
+/// # Deltas
+///
+/// A jar keeps track of any modifications made to it over time. The
+/// modifications are recorded as cookies. The modifications can be retrieved
+/// via [delta](#method.delta). Any new `Cookie` added to a jar via `add`
+/// results in the same `Cookie` appearing in the `delta`; cookies added via
+/// `add_original` do not count towards the delta. Any _original_ cookie that is
+/// removed from a jar results in a "removal" cookie appearing in the delta. A
+/// "removal" cookie is a cookie that a server sends so that the cookie is
+/// removed from the client's machine.
+///
+/// Deltas are typically used to create `Set-Cookie` headers corresponding to
+/// the changes made to a cookie jar over a period of time.
+///
+/// ```rust
+/// # use cookie::{Cookie, CookieJar};
+/// let mut jar = CookieJar::new();
+///
+/// // original cookies don't affect the delta
+/// jar.add_original(Cookie::new("original", "value"));
+/// assert_eq!(jar.delta().count(), 0);
+///
+/// // new cookies result in an equivalent `Cookie` in the delta
+/// jar.add(Cookie::new("a", "one"));
+/// jar.add(Cookie::new("b", "two"));
+/// assert_eq!(jar.delta().count(), 2);
+///
+/// // removing an original cookie adds a "removal" cookie to the delta
+/// jar.remove(Cookie::named("original"));
+/// assert_eq!(jar.delta().count(), 3);
+///
+/// // removing a new cookie that was added removes that `Cookie` from the delta
+/// jar.remove(Cookie::named("a"));
+/// assert_eq!(jar.delta().count(), 2);
+/// ```
+#[derive(Default, Debug, Clone)]
+pub struct CookieJar {
+    original_cookies: HashSet<DeltaCookie>,
+    delta_cookies: HashSet<DeltaCookie>,
 }
 
-enum Flavor<'a> {
-    Child(Child<'a>),
-    Root(Root),
-}
-
-struct Child<'a> {
-    parent: &'a CookieJar<'a>,
-    read: Read,
-    write: Write,
-}
-
-type Read = fn(&Root, Cookie<'static>) -> Option<Cookie<'static>>;
-type Write = fn(&Root, Cookie<'static>) -> Cookie<'static>;
-
-struct Root {
-    map: RefCell<HashMap<Cow<'static, str>, Cookie<'static>>>,
-    new_cookies: RefCell<HashSet<Cow<'static, str>>>,
-    removed_cookies: RefCell<HashSet<Cow<'static, str>>>,
-    _key: secure::SigningKey,
-}
-
-/// Iterator over the cookies in a cookie jar
-pub struct Iter<'a> {
-    jar: &'a CookieJar<'a>,
-    keys: Vec<Cow<'static, str>>,
-}
-
-impl<'a> CookieJar<'a> {
-    /// Creates a new empty cookie jar with the given signing key.
+impl CookieJar {
+    /// Creates an empty cookie jar.
     ///
-    /// The given key is used to sign cookies in the signed cookie jar. If
-    /// signed cookies aren't used then you can pass an empty array.
-    pub fn new(key: &[u8]) -> CookieJar<'static> {
-        CookieJar {
-            flavor: Flavor::Root(Root {
-                map: RefCell::new(HashMap::new()),
-                new_cookies: RefCell::new(HashSet::new()),
-                removed_cookies: RefCell::new(HashSet::new()),
-                _key: secure::prepare_key(key),
-            }),
-        }
+    /// # Example
+    ///
+    /// ```rust
+    /// use cookie::CookieJar;
+    ///
+    /// let jar = CookieJar::new();
+    /// assert_eq!(jar.iter().count(), 0);
+    /// ```
+    pub fn new() -> CookieJar {
+        CookieJar::default()
     }
 
-    fn root<'b>(&'b self) -> &'b Root {
-        let mut cur = self;
-        loop {
-            match cur.flavor {
-                Flavor::Child(ref child) => cur = child.parent,
-                Flavor::Root(ref me) => return me,
-            }
-        }
+    /// Finds a `Cookie` inside this jar with the name `name` and returns a
+    /// borrow to the cookie if it is found. Otherwise returns `None`.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use cookie::{CookieJar, Cookie};
+    ///
+    /// let mut jar = CookieJar::new();
+    /// assert!(jar.find("name").is_none());
+    ///
+    /// jar.add(Cookie::new("name", "value"));
+    /// assert_eq!(jar.find("name").map(|c| c.value()), Some("value"));
+    /// ```
+    pub fn find<'a>(&'a self, name: &str) -> Option<&'a Cookie<'static>> {
+        self.delta_cookies
+            .get(name)
+            .or_else(|| self.original_cookies.get(name))
+            .and_then(|c| if !c.removed { Some(&c.cookie) } else { None })
     }
 
-    /// Adds an original cookie from a request.
+    /// Adds an "original" `cookie` to this jar. Adding an original cookie does
+    /// not affect the [delta](#method.delta) computation. This method is
+    /// intended to be used to seed the cookie jar with cookies received from a
+    /// client's HTTP message.
     ///
-    /// This method only works on the root cookie jar and is not intended for
-    /// use during the lifetime of a request, it is intended to initialize a
-    /// cookie jar from an incoming request.
+    /// For accurate `delta` computations, this method should not be called
+    /// after calling `remove`.
     ///
-    /// # Panics
+    /// # Example
     ///
-    /// Panics if `self` is not the root cookie jar.
+    /// ```rust
+    /// use cookie::{CookieJar, Cookie};
+    ///
+    /// let mut jar = CookieJar::new();
+    /// jar.add_original(Cookie::new("name", "value"));
+    /// jar.add_original(Cookie::new("second", "two"));
+    ///
+    /// assert_eq!(jar.find("name").map(|c| c.value()), Some("value"));
+    /// assert_eq!(jar.find("second").map(|c| c.value()), Some("two"));
+    /// assert_eq!(jar.iter().count(), 2);
+    /// assert_eq!(jar.delta().count(), 0);
+    /// ```
     pub fn add_original(&mut self, cookie: Cookie<'static>) {
-        match self.flavor {
-            Flavor::Child(..) => panic!("can't add an original cookie to a child jar!"),
-            Flavor::Root(ref mut root) => {
-                let name = cookie.name().to_string();
-                root.map.borrow_mut().insert(name.into(), cookie);
-            }
-        }
+        self.original_cookies.replace(DeltaCookie::added(cookie));
     }
 
-    /// Adds a new cookie to this cookie jar.
-    ///
-    /// If this jar is a child cookie jar, this will walk up the chain of
-    /// borrowed jars, modifying the cookie as it goes along.
-    pub fn add(&self, mut cookie: Cookie<'static>) {
-        let mut cur = self;
-        let root = self.root();
-        loop {
-            match cur.flavor {
-                Flavor::Child(ref child) => {
-                    cookie = (child.write)(root, cookie);
-                    cur = child.parent;
-                }
-                Flavor::Root(..) => break,
-            }
-        }
-        let name = cookie.name().to_string();
-        root.map.borrow_mut().insert(name.clone().into(), cookie);
-        root.removed_cookies.borrow_mut().remove(&*name);
-        root.new_cookies.borrow_mut().insert(name.into());
-    }
-
-    /// Removes a cookie from this cookie jar.
-    pub fn remove<N: Into<Cow<'static, str>>>(&self, cookie_name: N) {
-        let root = self.root();
-        let name = cookie_name.into();
-        root.map.borrow_mut().remove(&name);
-        root.new_cookies.borrow_mut().remove(&name);
-        root.removed_cookies.borrow_mut().insert(name);
-    }
-
-    /// Clears all cookies from this cookie jar.
-    pub fn clear(&self) {
-        let root = self.root();
-        let all_cookies: Vec<_> = root.map
-            .borrow()
-            .keys()
-            .map(|n| n.to_owned())
-            .collect();
-
-        root.map.borrow_mut().clear();
-        root.new_cookies.borrow_mut().clear();
-        root.removed_cookies.borrow_mut().extend(all_cookies);
-    }
-
-    /// Finds a cookie inside of this cookie jar.
-    ///
-    /// The cookie is subject to modification by any of the child cookie jars
-    /// that are currently borrowed. A copy of the cookie is returned.
-    pub fn find(&self, name: &str) -> Option<Cookie<'static>> {
-        let root = self.root();
-        if root.removed_cookies.borrow().contains(name) {
-            return None;
-        }
-        root.map.borrow().get(name).and_then(|c| self.try_read(root, c.clone()))
-    }
-
-    /// Creates a child signed cookie jar.
-    ///
-    /// All cookies read from the child jar will require a valid signature and
-    /// all cookies written will be signed automatically.
-    ///
-    /// This API is only available when the `secure` feature is enabled on this
-    /// crate.
+    /// Adds `cookie` to this jar.
     ///
     /// # Example
     ///
     /// ```rust
-    /// # use cookie::{Cookie, CookieJar};
-    /// let c = CookieJar::new(b"f8f9eaf1ecdedff5e5b749c58115441e");
+    /// use cookie::{CookieJar, Cookie};
     ///
-    /// // Add a signed cookie to the jar
-    /// c.signed().add(Cookie::new("key".to_string(), "value".to_string()));
+    /// let mut jar = CookieJar::new();
+    /// jar.add(Cookie::new("name", "value"));
+    /// jar.add(Cookie::new("second", "two"));
     ///
-    /// // Add a permanently signed cookie to the jar
-    /// c.permanent().signed()
-    ///  .add(Cookie::new("key".to_string(), "value".to_string()));
+    /// assert_eq!(jar.find("name").map(|c| c.value()), Some("value"));
+    /// assert_eq!(jar.find("second").map(|c| c.value()), Some("two"));
+    /// assert_eq!(jar.iter().count(), 2);
+    /// assert_eq!(jar.delta().count(), 2);
     /// ```
-    #[cfg(feature = "secure")]
-    pub fn signed<'b>(&'b self) -> CookieJar<'b> {
-        return CookieJar {
-            flavor: Flavor::Child(Child {
-                parent: self,
-                read: verify,
-                write: sign,
-            }),
-        };
+    pub fn add(&mut self, cookie: Cookie<'static>) {
+        self.delta_cookies.replace(DeltaCookie::added(cookie));
+    }
 
-        fn verify(root: &Root, cookie: Cookie<'static>) -> Option<Cookie<'static>> {
-            secure::verify(&root._key, cookie)
+    /// Removes `cookie` from this jar. If an _original_ cookie with the same
+    /// name as `cookie` is present in the jar, a _removal_ cookie will be
+    /// present in the `delta` computation. To properly generate the removal
+    /// cookie, `cookie` must contain the same `path` and `domain` as the cookie
+    /// that was initially set.
+    ///
+    /// A "removal" cookie is a cookie that has the same name as the original
+    /// cookie but has an empty value, a max-age of 0, and an expiration date
+    /// far in the past.
+    ///
+    /// # Example
+    ///
+    /// Removing an _original_ cookie results in a _removal_ cookie:
+    ///
+    /// ```rust
+    /// # extern crate cookie;
+    /// extern crate time;
+    ///
+    /// use cookie::{CookieJar, Cookie};
+    /// use time::Duration;
+    ///
+    /// # fn main() {
+    /// let mut jar = CookieJar::new();
+    ///
+    /// // Assume this cookie originally had a path of "/" and domain of "a.b".
+    /// jar.add_original(Cookie::new("name", "value"));
+    ///
+    /// // If the path and domain were set, they must be provided to `remove`.
+    /// jar.remove(Cookie::build("name", "").path("/").domain("a.b").finish());
+    ///
+    /// // The delta will contain the removal cookie.
+    /// let delta: Vec<_> = jar.delta().collect();
+    /// assert_eq!(delta.len(), 1);
+    /// assert_eq!(delta[0].name(), "name");
+    /// assert_eq!(delta[0].max_age(), Some(Duration::seconds(0)));
+    /// # }
+    /// ```
+    ///
+    /// Removing a new cookie does not result in a _removal_ cookie:
+    ///
+    /// ```rust
+    /// use cookie::{CookieJar, Cookie};
+    ///
+    /// let mut jar = CookieJar::new();
+    /// jar.add(Cookie::new("name", "value"));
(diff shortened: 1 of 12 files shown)


src/jar.rs
```diff
@@ -1,540 +1,365 @@
...
+    /// ```rust
+    /// use cookie::{CookieJar, Cookie};
+    ///
+    /// let mut jar = CookieJar::new();
+    /// assert!(jar.find("name").is_none());
+    ///
+    /// jar.add(Cookie::new("name", "value"));
+    /// assert_eq!(jar.find("name").map(|c| c.value()), Some("value"));
+    /// ```
+    pub fn find<'a>(&'a self, name: &str) -> Option<&'a Cookie<'static>> {
```
alexcrichton: I think that the `'a` here can be elided, right?

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+    fn unseal(&self, value: &str) -> Result<String, &'static str> {
+        let (nonce_s, sealed_s) = value.split_at(BASE64_NONCE_LEN);
+        let nonce = nonce_s.from_base64().map_err(|_| "bad nonce base64")?;
+        let mut sealed = sealed_s.from_base64().map_err(|_| "bad sealed base64")?;
+        let key = OpeningKey::new(ALGO, self.key).expect("opening key");
+
+        let out_len = open_in_place(&key, &nonce, 0, &mut sealed, &[])
+            .map_err(|_| "invalid key/nonce/value: bad seal")?;
+
+        unsafe { sealed.set_len(out_len); }
```
alexcrichton: In the interest of removing `unsafe` in security-sensitive portions, could this just use `truncate`? It could always be switched back if it turns out to be a problem, of course.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use secure::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
+
+use {Cookie, CookieJar};
+
+// Keep these in sync, and keep the key len synced with the `private` docs.
+static ALGO: &'static Algorithm = &AES_256_GCM;
+const KEY_LEN: usize = 32;
+const NONCE_LEN: usize = 12;
+const BASE64_NONCE_LEN: usize = 16;
```
alexcrichton: Could this be expressed as a function of `NONCE_LEN`?

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+use secure::ring::rand::SystemRandom;
+
+use secure::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
+
+use {Cookie, CookieJar};
+
+// Keep these in sync, and keep the key len synced with the `private` docs.
+static ALGO: &'static Algorithm = &AES_256_GCM;
+const KEY_LEN: usize = 32;
+const NONCE_LEN: usize = 12;
```
alexcrichton: Is there perhaps a reference to add for why 12 was chosen?

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// random.
+    ///
+    /// Any modifications to the child jar will be reflected on the parent jar,
+    /// and any retrievals from the child jar will be made from the parent jar.
+    ///
+    /// This trait is only available when the `secure` feature is enabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `key` is not exactly 64 bytes long.
```
alexcrichton: (similar concerns about usability with the 32-byte restriction above, just makign a note)

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// # let key: Vec<_> = (0..64).collect();
+    /// let mut jar = CookieJar::new();
+    /// jar.signed(&key).add(Cookie::new("name", "value"));
+    ///
+    /// assert_ne!(jar.find("name").unwrap().value(), "value");
+    /// assert!(jar.find("name").unwrap().value().contains("value"));
+    /// assert_eq!(jar.signed(&key).find("name").unwrap().value(), "value");
+    /// ```
+    pub fn add(&mut self, mut cookie: Cookie<'static>) {
+        let key = SigningKey::new(DIGEST, self.key);
```
alexcrichton: Since this is constant, is there a downside to creating this then the `Signed` jar was created? That'd help avoid the `'k` lifetime parameter I think as well.

alexcrichton: Oh I think this'd apply to `Private` as well

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// let mut jar = CookieJar::new();
+    /// let mut signed_jar = jar.signed(&key);
+    ///
+    /// signed_jar.add(Cookie::new("name", "value"));
+    /// assert!(signed_jar.find("name").is_some());
+    ///
+    /// signed_jar.remove(Cookie::named("name"));
+    /// assert!(signed_jar.find("name").is_none());
+    /// ```
+    pub fn remove(&mut self, cookie: Cookie<'static>) {
```
alexcrichton: Could this (and other remove methods) perhaps return `Option<Cookie<'static>>`? (the value that was actually removed)

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
+use secure::ring::digest::{SHA256, Algorithm};
+use secure::ring::hmac::{SigningKey, sign, verify_with_own_key as verify};
+
+use secure::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
+
+use {Cookie, CookieJar};
+
+// Keep these three in sync, and keep the key len synced with the `signed` docs.
+static DIGEST: &'static Algorithm = &SHA256;
+const BASE64_DIGEST_LEN: usize = 44;
```
alexcrichton: Could this be a function of `KEY_LEN`?

alexcrichton:
Also cc @briansmith (usage of *ring* here), although to me at least it all looked great! (and ergonomic to use)

src/jar.rs
SergioBenitez: Yep, the `'a` can be elided. Yeah, `get` sounds good to me. I'm against implementing `Index` since the contents of the `CookieJar` are usually determined by some opaque actor (i.e, a web framework), so you don't know what's in the jar. As a result, the index is likely to panic often, thus being error prone. We could return an `Option` from the `Index` `impl`, but then it'll stray from convention.

src/jar.rs
```diff
@@ -1,540 +1,365 @@
...
-            now.tm_year -= 1;
-            c.set_expires(now);
-            ret.push(c);
-        }
-
-        let map = root.map.borrow();
-        for cookie in root.new_cookies.borrow().iter() {
-            ret.push(map.get(cookie).unwrap().clone());
+    /// Removes all cookies from this cookie jar.
+    #[deprecated(since = "0.7.0", note = "calling this method may not remove \
```
SergioBenitez: I wanted to deprecate it to smooth transition because I feel that this method may actually be used in the wild. I don't know if that's true or not, however.

src/secure/private.rs
SergioBenitez: The nonce size is determined by the AEAD algorithm, here, `12` for a 96-bit nonce.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// let mut private_jar = jar.private(&key);
+    /// assert!(private_jar.find("name").is_none());
+    ///
+    /// private_jar.add(Cookie::new("name", "value"));
+    /// assert_eq!(private_jar.find("name").unwrap().value(), "value");
+    /// ```
+    pub fn find(&self, name: &str) -> Option<Cookie<'static>> {
+        if let Some(cookie_ref) = self.parent.find(name) {
+            let mut cookie = cookie_ref.clone();
+            if cookie.value().len() <= BASE64_NONCE_LEN {
```
briansmith: Shouldn't this check be in `unseal()` to avoid a panic?

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// let mut jar = CookieJar::new();
+    /// jar.private(&key).add(Cookie::new("name", "value"));
+    ///
+    /// assert_ne!(jar.find("name").unwrap().value(), "value");
+    /// assert_eq!(jar.private(&key).find("name").unwrap().value(), "value");
+    /// ```
+    pub fn add(&mut self, mut cookie: Cookie<'static>) {
+        // Generate the nonce.
+        let mut nonce = [0; NONCE_LEN];
+        SystemRandom::new().fill(&mut nonce).expect("couldn't randomly fill nonce");
```
briansmith: At a minimum, please document the fact that using a random nonce this short is not the best thing.

The nonce is only 96 bits. That means there is a 2<sup>48</sup> chance of collision even assuming the RNG is perfect. That's usually too low for nonce generation, which is why I don't recommend using regular AEAD algorithms for this kind of application, and also why there is no `SealingKey::open_in_place_with_own_key()`. In https://github.com/briansmith/ring/issues/411 and other issues, we're tracking adding better algorithms to *ring* to support this kind of use case.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        // Create the `SealingKey` structure.
+        let key = SealingKey::new(ALGO, self.key).expect("sealing key creation");
+
+        // Setup the input and output for the sealing operation.
+        let overhead = ALGO.max_overhead_len();
```
briansmith: Why not make `in_out` contain nonce+ciphertext+tag (tag == overhead), and then base64-encode the whole thing in one shot? It seems like the resulting value would be slightly more efficiently encoded, and you'd save some mallocs in this function and in `unseal()`, I think.

Then...

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+        let mut in_out = {
+            let cookie_val = cookie.value().as_bytes();
+            let mut in_out = vec![0; cookie_val.len() + overhead];
```
briansmith: ...you could change this to:

```rust
let mut in_out = vec![0; ALGO.nonce_len() + cookie_val.len() + overhead];
let (nonce, in_out) = in_out.split_at_mut(ALGO.nonce_len());
rng.fill(nonce).expect("...");
```
Notably, this would avoid you needing to define your own const `NONCE_LEN`.

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+/// A _signed_ child jar signs all the cookies added to it and verifies cookies
+/// retrieved from it. Any cookies stored in a `SignedJar` are assured integrity
+/// and authenticity. In other words, clients cannot tamper with the contents of
+/// a cookie nor can they fabricate cookie values, but the data is visible in
+/// plaintext.
+///
+/// This type is only available when the `secure` feature is enabled.
+pub struct SignedJar<'a, 'k> {
+    parent: &'a mut CookieJar,
+    key: &'k [u8]
```
briansmith: Why not `key: SigningKey`? That's how *ring* is designed to be used. Note in particular that `SigningKey::new()` does part of the HMAC computation, so subsequent uses of the same `SigningKey` are cheaper.

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+
+impl<'a, 'k> SignedJar<'a, 'k> {
+    /// Given a signed value `str` where the signature is prepended to `value`,
+    /// verifies the signed value and returns it. If there's a problem, returns
+    /// an `Err` with a string describing the issue.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `value.len()` < BASE64_DIGEST_LEN.
+    fn verify(&self, cookie_value: &str) -> Result<String, &'static str> {
```
briansmith: Why panic when the value is too short, instead of returning `Err`?

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// let mut signed_jar = jar.signed(&key);
+    /// assert!(signed_jar.find("name").is_none());
+    ///
+    /// signed_jar.add(Cookie::new("name", "value"));
+    /// assert_eq!(signed_jar.find("name").unwrap().value(), "value");
+    /// ```
+    pub fn find(&self, name: &str) -> Option<Cookie<'static>> {
+        if let Some(cookie_ref) = self.parent.find(name) {
+            let mut cookie = cookie_ref.clone();
+            if cookie.value().len() < BASE64_DIGEST_LEN {
```
briansmith: IMO this check should be in `verify()`.

src/secure/private.rs
SergioBenitez: Yes, it's:

```rust
const BASE64_NONCE_LEN: usize = (((NONCE_LEN * 4) / 3) + 3) & !0b11;
```

Looks a bit unwieldily, though, which is why I just precomputed the number.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// _must_ be cryptographically random.
+    ///
+    /// Any modifications to the child jar will be reflected on the parent jar,
+    /// and any retrievals from the child jar will be made from the parent jar.
+    ///
+    /// This trait is only available when the `secure` feature is enabled.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `key` is not exactly 32 bytes long.
```
SergioBenitez: We could accept keys of length _greater_ than 32 bytes, but accepting and padding anything below that weakens security if the attacker knows the size of the key the user is using.

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+static DIGEST: &'static Algorithm = &SHA256;
+const BASE64_DIGEST_LEN: usize = 44;
+const KEY_LEN: usize = 64;
+
+/// Extends `CookieJar` with a `signed` method to retrieve a signed child jar.
+pub trait Signed<'a, 'k> {
+    /// Returns a `SignedJar` with `self` as its parent jar using the key `key`
+    /// to sign/verify cookies added/retrieved from the child jar. The key must
+    /// be exactly 64 bytes. For security, the key _must_ be cryptographically
+    /// random.
```
briansmith: Also, the caller should ensure that different and unrelated keys are used for `Signed` and `Private`, since the same key shouldn't be used by two algorithms.

src/secure/private.rs
SergioBenitez: Will `truncate` result in an allocation? It's not clear from the documentation. Here I'm assuming `ring`'s return value is correct, but I should be checking that `sealed.len() >= out_len`.

src/secure/signed.rs
SergioBenitez: No, this is a function of `DIGEST_LEN`, which isn't declared but would be `256 / 8 = 32`. The function would the same as for `BASE64_NONCE_LEN`.

src/secure/signed.rs
SergioBenitez: The only downside is that we might have to include the lifetime in the future if `SigningKey` changes or if we use a different library that doesn't store the key directly. In other words, we lose flexibility to swap out `ring` easily.

src/secure/private.rs
briansmith: The problem with taking a variable-length key and hashing it is that many people will use a password instead of a key. And a hash of a password isn't a strong key. Taking a fixed length of bytes as a key encourages people to read the docs and not pass a password instead of a real key.

If you are worried about the implications of changing algorithms in the future, then I recommend that you provide a function that returns the currently-required key length.

And/or, you could design the API such that the user only has to supply one master (256-bit) key, and then different keys for HMAC-signing and sealing are derived (e.g. using `ring::hkdf`) from that master key. That's what I recommend, because then it is less likely that the user will use related keys for HMAC signing and for sealing.

And/or, you could replace the use of HMAC for sign-only operations with the use of `seal_in_place()`, where the data to be signed-but-not-encrypted is the `ad` and the plaintext is empty. This will sign with GHASH (GCM) instead of HMAC in the case of AES-GCM. Then it is OK to use the same key for both signing-only and sign-and-encrypt operation.

src/secure/signed.rs
SergioBenitez: Well, we reuse the cookie removed from `original_cookies`, so we'd have to clone there. If the cookie was previously added via `add`, we could just return.

I'd rather not `clone` unnecessary. We could return a `bool` though.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+use {Cookie, CookieJar};
+
+// Keep these in sync, and keep the key len synced with the `private` docs.
+static ALGO: &'static Algorithm = &AES_256_GCM;
+const KEY_LEN: usize = 32;
+const NONCE_LEN: usize = 12;
+const BASE64_NONCE_LEN: usize = 16;
+
+/// Extends `CookieJar` with a `private` method to retrieve a private child jar.
+pub trait Private<'a, 'k> {
```
SergioBenitez: I don't...think so. Let me think about that for a bit and change it if I can't think of any reason.

src/secure/signed.rs
briansmith: It's approximately `DIGEST.output_len` * 5 / 4`, or whatever the overhead calculation for base64 is.

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
+use secure::ring::digest::{SHA256, Algorithm};
+use secure::ring::hmac::{SigningKey, sign, verify_with_own_key as verify};
+
+use secure::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
+
+use {Cookie, CookieJar};
+
+// Keep these three in sync, and keep the key len synced with the `signed` docs.
+static DIGEST: &'static Algorithm = &SHA256;
```
briansmith: IMO, `SIGNING_HMAC_DIGEST_ALG` or similar is a better name for this. In particular, it is strange that we have `DIGEST` here and then `ALGO` for the AEAD algorithm.

src/secure/private.rs
briansmith: It's `ALGO.nonce_len()`.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+use secure::ring::aead::{OpeningKey, SealingKey};
+use secure::ring::rand::SystemRandom;
+
+use secure::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
+
+use {Cookie, CookieJar};
+
+// Keep these in sync, and keep the key len synced with the `private` docs.
+static ALGO: &'static Algorithm = &AES_256_GCM;
+const KEY_LEN: usize = 32;
```
briansmith: Maybe document that this is `ALGO.key_len()`.

src/secure/signed.rs
```diff
@@ -0,0 +1,206 @@
...
+const BASE64_DIGEST_LEN: usize = 44;
+const KEY_LEN: usize = 64;
```
briansmith: Why did you choose the digest algorithm block size instead of the digest algorithm output size?

src/secure/private.rs
briansmith: I don't think you need to check that `seeled.len() >= out_len` as *ring* does guarantee that. It will be clearer in an upcoming version of *ring*, where we will change the return value to be a slice of the `ciphertext` parameter.

src/secure/signed.rs
briansmith: FWIW, in the upcoming, not-yet-released version of *ring*, I've added a new function to generate a serializable HMAC key; see https://github.com/briansmith/ring/commit/f0d476018ce5d99926ac38b2c95f11fefe820f83, based on this review.

briansmith: Based on this review, in the upcoming version of *ring*, I added a new API to help people generate serializable HMAC keys of a good length. See
https://github.com/briansmith/ring/commit/f0d476018ce5d99926ac38b2c95f11fefe820f83. Using the block size isn't wrong though.

src/secure/private.rs
SergioBenitez: Sure, I wasn't trying to imply that it doesn't indeed try to do that. What I mean is that in order for this particular `unsafe` to be guaranteed `safe`, that's the condition that must be true. It's nice to have the invariants be verified locally.

src/secure/signed.rs
SergioBenitez: See my comment on `BASE64_NONCE_LEN` for the true calculation.

src/secure/private.rs
SergioBenitez: I think the "keep in sync comment" implies this pretty directly.

src/secure/private.rs
```diff
@@ -0,0 +1,231 @@
...
+/// A _private_ child jar signs and encrypts all the cookies added to it and
+/// verifies and decrypts cookies retrieved from it. Any cookies stored in a
+/// `PrivateJar` are simultaneously assured confidentiality, integrity, and
+/// authenticity. In other words, clients cannot discover nor tamper with the
+/// contents of a cookie, nor can they fabricate cookie data.
+///
+/// This type is only available when the `secure` feature is enabled.
+pub struct PrivateJar<'a, 'k> {
+    parent: &'a mut CookieJar,
+    key: &'k [u8]
```
SergioBenitez: This ties us to `ring` pretty heavily. See my other comment about whether this tradeoff is worth it or not.

src/secure/private.rs
SergioBenitez: The docstring for `unseal` clarifies its intentions, and since it's used a single time, it doesn't really matter where this is.

src/secure/private.rs
SergioBenitez: I don't think this is something to be concerned about. AES-256-GCM is the industry standard AEAD, and while 96 bits of nonce isn't the best-possible-thing, I don't think it's reasonable to be concerned about a nonce collision occurring after 281,474,976,710,656 cookies have been encrypted.

In any case, I designed the code such that replacing the AEAD algorithm with something better is simply changing 3 trivial lines. Once there's something better, we should certainly move to that.

src/secure/signed.rs
SergioBenitez: Yeah, I toyed with the name a bit here. `SIGNING_HMAC_DIGEST_ALG` is just too long for no real gain. Maybe just `ALGO`, or `HMAC_DIGEST`?

src/secure/signed.rs
SergioBenitez: No great reason except that it makes brute-force-key searching harder. So while the security of any leaked, decrypted cookies via some collision doesn't increase, the probability that they've found the real key is greatly reduced. Likely doesn't matter in practice.

src/secure/signed.rs
SergioBenitez: Indeed. I just assumed users would know, but this is a bad assumption. I'll make a note.

src/secure/signed.rs
SergioBenitez: See my other note on this comment.

src/secure/signed.rs
SergioBenitez: Yeah, that's fair.

src/secure/signed.rs
SergioBenitez: Okay.

src/secure/private.rs
SergioBenitez: Awesome suggestion. Thank you!

SergioBenitez:
I _wish_ we could migrate to `std::time`, but unless I'm missing something, it doesn't have a concept of absolute calendar time such as [Tm](https://doc.rust-lang.org/time/time/struct.Tm.html). I've tried working with `chrono` before and unfortunately am not a fan. :(

On the `base64` front: I'm totally down for using something more focused, but what? I'd guess the [base64 crate](https://crates.io/crates/base64) would be the thing to use. Do you think it's worth it? `rustc-serialize is a fairly lightweight dependency, even though it includes much more functionality than we're using.

src/secure/private.rs
briansmith: > This ties us to ring pretty heavily. See my other comment about whether this tradeoff is worth it or not.

I don't think it more tightly couples the library to *ring* in any significant way because these fields are private and so can't be accessed outside the module/crate.

src/secure/private.rs
briansmith: > AES-256-GCM is the industry standard AEAD, and while 96 bits of nonce isn't the best-possible-thing, I don't think it's reasonable to be concerned about a nonce collision occurring after 281,474,976,710,656 cookies have been encrypted.

Here is what the actual AES-GCM standard, from NIST, says:

> Unless an implementation only uses 96-bit IVs that are generated by the deterministic construction: **The total number of invocations of the authenticated encryption function shall not exceed 2<sup>32</sup>, including all IV lengths and all instances of the authenticated encryption function with the given key.**

(The emphasis is NIST's, not mine.)

In other words, if you use random nonces with AES-GCM then you must not encrypt more than ~4 billion cookies with the same key before replacing the key, according to the standard.

src/secure/signed.rs
briansmith: > No great reason except that it makes brute-force-key searching harder. So while the security of any leaked, decrypted cookies via some collision doesn't increase, the probability that they've found the real key is greatly reduced. Likely doesn't matter in practice.

It's not a big deal. I just want to point out that if the difference between a 256-bit key and a 512-bit key matters here then SHA-256 probably isn't the right thing to use. In general I recommend people use 256-bit keys for SHA-256 and 512-bit keys for SHA-384 (and don't use SHA-512).

src/secure/signed.rs
briansmith: I think `HMAC_DIGEST` is better than `ALGO`.

src/secure/private.rs
```diff
@@ -0,0 +1,184 @@
...
+use secure::ring::rand::SystemRandom;
+
+use secure::rustc_serialize::base64::{ToBase64, FromBase64, STANDARD};
+
+use {Cookie, CookieJar};
+
+// Keep these in sync, and keep the key len synced with the `private` docs.
+static ALGO: &'static Algorithm = &AES_256_GCM;
+const KEY_LEN: usize = 32;
+const NONCE_LEN: usize = 12;
```
briansmith: I reviewed the whole PR and I think you don't need the `KEY_LEN` or `NONCE_LEN` constants at all. You can now always use `ALGO.key_len()` and `ALGO.nonce_len()` instead, AFAICT, since there's no place where you need them to be compile-time constants.

src/secure/signed.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Given a signed value `str` where the signature is prepended to `value`,
+    /// verifies the signed value and returns it. If there's a problem, returns
+    /// an `Err` with a string describing the issue.
+    fn verify(&self, cookie_value: &str) -> Result<String, &'static str> {
+        if cookie_value.len() < BASE64_DIGEST_LEN {
+            return Err("length of value is <= BASE64_DIGEST_LEN");
+        }
+
+        let (digest_str, value) = cookie_value.split_at(BASE64_DIGEST_LEN);
```
briansmith: If you base64-decode first and then split, you don't need to define `BASE64_DIGEST_LEN`, since you can use `HMAC_ALG.output_len` for the split index instead.

briansmith:
FWIW, I released *ring* 0.7 with a new API for the HMAC key length recommendation; see https://briansmith.org/rustdoc/ring/hmac/fn.recommended_key_len.html. Your choice is also correct, but if you use `recommended_key_len(HMAC_DIGEST_ALG)` instead, then you don't need to have a separate `KEY_LEN` constant.

src/secure/signed.rs
briansmith: > If you base64-decode first and then split, you don't need to define BASE64_DIGEST_LEN, since you can use HMAC_ALG.output_len for the split index instead.

Sorry, I just realized that it doesn't make sense to base64-encode the entire value since, presumably, the initial value was already header-safe and base64-encoding would be unnecessarily wasteful.

src/jar.rs
alexcrichton: Sure yeah, seems ok to not implement `Index` for now (certainly a conservative approach)

src/secure/private.rs
alexcrichton: FWIW no, `truncate` won't allocate

src/secure/signed.rs
alexcrichton: Oh I thought we already had the cookie on hand to return and we just dropped it somewhere. Reading again though it looks like the returned value is squirreled away somewhere else, so yeah let's just return a bool.

src/secure/private.rs
alexcrichton: I also like @briansmith's suggestion, but @SergioBenitez to leave the door open to having a slice of bytes in the future perhaps this could add a `PhantomData<&'k [u8]>`?

alexcrichton:
Thanks so much for the review @briansmith!

alexcrichton:
> I wish we could migrate to std::time, but unless I'm missing something, it doesn't have a concept of absolute calendar time such as Tm.

Yeah that's true, all we'd get is offsets from `SystemTime` and such. I'd love to drop `time` as a dep and leave heavy lifting (if necessary) to other crates, but the time may not yet be ripe to do so.

> On the base64 front: I'm totally down for using something more focused, but what? I'd guess the base64 crate would be the thing to use. Do you think it's worth it? rustc-serialize is a fairly lightweight dependency, even though it includes much more functionality than we're using.

Oh just kinda a random thought, I'd love to burn rustc-serialize to the ground but it's a private dep so we can change this at any time :)

src/secure/signed.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl<'a, 'k> SignedJar<'a, 'k> {
+    /// Creates a new child `SignedJar` with parent `parent` and key `key`. This
+    /// method is typically called indirectly via the `signed` method of
+    /// `CookieJar`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `key` is not exactly 64 bytes long.
+    #[doc(hidden)]
```
alexcrichton: Could this be a submodule of `jar` so the constructor doesn't need to be public to the world?

briansmith:
> On the base64 front: I'm totally down for using something more focused, but what? I'd guess the base64 crate would be the thing to use. Do you think it's worth it? rustc-serialize is a fairly lightweight dependency, even though it includes much more functionality than we're using.

In #78 I suggest using Z85 encoding instead of Base64 encoding. It's more space-efficient and just as compatible as Base64-encoding, AFAICT. Also there are two crates that implement Z85 already.

src/secure/private.rs
briansmith: > slice of bytes in the future perhaps this could add a PhantomData<&'k [u8]>?

Is it really a good idea to reference the key as a slice anyway? In typical use, the user would need to create a wrapper around the `PrivateJar` and key storage just to accomodate the key being a slice. IMO, even if my suggestion isn't taken, the key should be copied into a known value in the `PrivateJar`.

alexcrichton:
@briansmith oh oops sorry! I was away all weekend and been digging backwards through email so I ended up seeing that later :). In any case, sounds like a great idea to me!

src/secure/private.rs
SergioBenitez: @briansmith That's exactly what I was thinking: in any case, we could just copy the key into the structure. I've removed the lifetime.

src/secure/signed.rs
SergioBenitez: You're right. I'll just use 32-byte keys.

src/jar.rs
```diff
@@ -1,637 +1,608 @@
...
-        };
-
-        fn verify(root: &Root, cookie: Cookie<'static>) -> Option<Cookie<'static>> {
-            secure::verify(&root._key, cookie)
-        }
-
-        fn sign(root: &Root, cookie: Cookie<'static>) -> Cookie<'static> {
-            secure::sign(&root._key, cookie)
-        }
+    pub fn private<'a, 'k>(&'a mut self, key: &[u8]) -> PrivateJar<'a> {
```
alexcrichton: This `'k` is now unused I think

alexcrichton: (and I think the `'a` can be elided)

src/secure/private.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// jar.private(&key).add(Cookie::new("name", "value"));
+    ///
+    /// assert_ne!(jar.get("name").unwrap().value(), "value");
+    /// assert_eq!(jar.private(&key).get("name").unwrap().value(), "value");
+    /// ```
+    pub fn add(&mut self, mut cookie: Cookie<'static>) {
+        let mut data;
+        let output_len = {
+            // Create the `SealingKey` structure.
+            let key = SealingKey::new(ALGO, &self.key).expect("sealing key creation");
```
alexcrichton: Could this be stored in the `PrivateJar` struct itself?

alexcrichton:
Ok so API-wise I think we're all set here modulo the key length question. I'd like to not have a restriction that the keys passed in are a precise length because that seems (a) difficult to change over time and (b) surprising in terms of trying to match it.

It sounds like hashing the input is a bad idea, but could we perhaps say that you need to pass in at *least* X bytes of data and we use slices within that for various keys? That way we can also provide a sample command for generating the keys to pass in perhaps.

briansmith:
> I'd like to not have a restriction that the keys passed in are a precise length because that seems (a) difficult to change over time and (b) surprising in terms of trying to match it.

Keep in mind that if/when the algorithms change, the keys need to change too. For example, let's say you switch AES-256-GCM to ChaCha20-Poly1305. Even though the key size is the same, it isn't safe to use the same key (or, rather, we assume it is unsafe and nobody has proven otherwise, nor will anybody ever do so).

Thus, if you want to have some flexibility then you need to specify, *at least*, a minimum key length, e.g. 256 bits. Then you can use HKDF, (see [`ring::hkdf::extract_and_expand`](https://briansmith.org/rustdoc/ring/hkdf/fn.extract_and_expand.html)) to derive the keys you need. The `info` parameter needs to be something that unambiguously identifies the algorithm, usage, and context, e.g. `"HTTP-cookie-encrypt-and-auth;AES-256-GCM;v1.0;shared-across-origin:"+origin` for the AEAD key and `"HTTP-cookie-auth-only;HMAC-SHA2-256;v1.0;shared-across-origin: " + origin` for the HMAC key, where `origin` is the origin of the cookie.

Then whenever you change anything related to the crypto in cookie-rs, you need to change the those labels. Also note that with this scheme, different origins will always have different keys. For this latter reason alone, it may be a good idea to move to the derived-key scheme.

However, still, I recommend that instead of specifying a minimum key length, you instead specify a fixed key length. In general we (crypto people) don't like variables, especially regarding key lengths. (We now generally regard the way HMAC's key was specified as a mistake.)

SergioBenitez:
@alexcrichton I'm not sure I understand what you mean by "sample command for generating the keys to pass in". Do you mean like a shell command? If so, how would that be any different if we had a fixed-key length? And if you mean some example Rust code, then you should consider that most of these keys will be fixed. That is, they won't be generated at runtime. This is because (especially in production) you generally want to accept cookies that were signed the last time you started your application or by other instances of that application. As a result, you need one, fixed key.

In general, I'm in agreement with @briansmith here, as evidenced by my choice of a fixed-key length to begin with.

alexcrichton:
@briansmith that sounds like a great idea! (sorry I'm pretty unfamiliar with hkdf). I'd be totally fine though requiring a fixed key length if we can use that to derive further keys (even if we change underlying algorithms)

@SergioBenitez eh I basically just mean as a new users I see this key restriction or these random constants pass in and I'm not sure how to do so myself. Just some docs for "here's how to create your own key then pass it in".

SergioBenitez:
@alexcrichton I think allowing variable sized keys or extracting keys of different lengths from one key is something that should be built on-top of `cookie` but not done by `cookie` itself. I think requiring a fixed-size key is the safest thing we can do. What's more, I don't see the 32-byte key size changing any time soon. The only algorithm I can see us switching to anywhere in the vicinity of the near future is `AES-256-GCM-SIV` which, if I'm not mistaken, also requires 32-byte keys.

As far as how to get these random constants...the safest way (and the way Rocket will do this) is to encode the key in a text-safe format (base64 in Rocket) and then get the byte representation of the key. It's trivial to generate a safe 32-byte random key in this way: `openssl rand -base64 32`.

briansmith:
> don't see the 32-byte key size changing any time soon. The only algorithm I can see us switching to anywhere in the vicinity of the near future is AES-256-GCM-SIV which, if I'm not mistaken, also requires 32-byte keys.

But, you shouldn't use the same key for different algorithms. So if there's any possibility that the algorithm will change (and I think there is), it is better to derive the keys from the master key.

> As far as how to get these random constants...the safest way (and the way Rocket will do this) is to encode the key in a text-safe format (base64 in Rocket) and then get the byte representation of the key. It's trivial to generate a safe 32-byte random key in this way: openssl rand -base64 32.

Yes, and this is why variable-length inputs are bad, because some callers are likely to pass in the base64-encoded key bytes, instead of decoding the base64 first. Using base64-encoded bytes as a key, without decoding them, is bad.

SergioBenitez:
> But, you shouldn't use the same key for different algorithms. So if there's any possibility that the algorithm will change (and I think there is), it is better to derive the keys from the master key.

How do we defend against this with `hkdf`? Use a different salt each time we change the algorithm?

Running `hkdf` isn't free. Again, I think if a library on-top of `cookie` wants to use `hkdf`, then so be it, but it's not something I think `cookie` should handle.

> Yes, and this is why variable-length inputs are bad, because some callers are likely to pass in the base64-encoded key bytes, instead of decoding the base64 first. Using base64-encoded bytes as a key, without decoding them, is bad.

Indeed.

briansmith:
> Running hkdf isn't free. Again, I think if a library on-top of cookie wants to use hkdf, then so be it, but it's not something I think cookie should handle.

Yes, I explained in https://github.com/alexcrichton/cookie-rs/pull/76#issuecomment-281863128.

I don't want to get too hung up in providing advice here, so I won't debate it further.

alexcrichton:
@SergioBenitez lets use hkdf. "This probably won't change soon" sounds like "when this changes it'll be too difficult to change when we really need to". Let's cover our bases and start out conservatively? If this really has a perf impact or something like that we can reevaluate.

SergioBenitez:
Okay. Now using HKDF via a `Key` struct. I quite like this approach.

This should be all!

alexcrichton:
👍

alexcrichton:
@SergioBenitez do you want to do the release on this one?

SergioBenitez:
Released! :)

PR #77, opened 2017-02-15
Added some new methods to get strings with longer lifetimes

golddranks:
Added some new methods to get strings with longer lifetimes if they are in the backing storage of the Cookie.

This is still WIP; the names are provisional, no tests implemented and the docstrings aren't up to date.

What do you think about this idea? Certainly less invasive than my earlier experiment. If it seems fine, I'll finish this up.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,14 +89,6 @@ enum CookieStr {
 }
 
 impl CookieStr {
-    /// Whether this string is derived from indexes or not.
-    fn is_indexed(&self) -> bool {
-        match *self {
-            CookieStr::Indexed(..) => true,
-            CookieStr::Concrete(..) => false,
-        }
-    }
-
     /// Retrieves the string `self` corresponds to. If `self` is derived from
     /// indexes, the corresponding subslice of `string` is returned. Otherwise,
     /// the concrete string is returned.
@@ -105,13 +97,25 @@ impl CookieStr {
     ///
     /// Panics if `self` is an indexed string and `string` is None.
     fn to_str<'s>(&'s self, string: Option<&'s Cow<str>>) -> &'s str {
-        if self.is_indexed() && string.is_none() {
-            panic!("Cannot convert indexed str to str without base string!")
-        }
-
         match *self {
-            CookieStr::Indexed(i, j) => &string.unwrap()[i..j],
+            CookieStr::Indexed(i, j) => {
+                let s = string.expect("`Some` base string must exist when \
+                    converting indexed str to str! (This is a module invariant.)");
+                &s[i..j]
+            },
             CookieStr::Concrete(ref cstr) => &*cstr,
+        }
+    }
+
+    fn to_raw_str<'s, 'c: 's>(&'s self, string: &'s Cow<'c, str>) -> Option<&'c str> {
+        match *self {
+            CookieStr::Indexed(i, j) => {
+                match *string {
+                    Cow::Borrowed(s) => Some(&s[i..j]),
+                    Cow::Owned(_) => None,
+                }
+            },
+            CookieStr::Concrete(_) => None,
         }
     }
 }
@@ -640,6 +644,134 @@ impl<'c> Cookie<'c> {
 
         Ok(())
     }
+
+    /// Returns the name of `self` as a string slice of the raw string `self`
+    /// was originally parsed from. If `self` was not originally parsed from a
+    /// raw string, returns `None`.
+    ///
+    /// This method differs from [name](#method.name) in that it returns a
+    /// string with the same lifetime as the originally parsed string. This
+    /// lifetime may outlive `self`. If a longer lifetime is not required, or
+    /// you're unsure if you need a longer lifetime, use [name](#method.name).
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let cookie_string = format!("{}={}", "foo", "bar");
+    ///
+    /// // `c` will be dropped at the end of the scope, but `name` will live on
+    /// let name = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.name_raw()
+    /// };
+    ///
+    /// assert_eq!(name, Some("foo"));
+    /// ```
+    #[inline]
+    pub fn name_raw(&self) -> Option<&'c str> {
+        self.cookie_string.as_ref()
+            .and_then(|s| self.name.to_raw_str(s))
+    }
+
+    /// Returns the value of `self` as a string slice of the raw string `self`
+    /// was originally parsed from. If `self` was not originally parsed from a
+    /// raw string, returns `None`.
+    ///
+    /// This method differs from [value](#method.value) in that it returns a
+    /// string with the same lifetime as the originally parsed string. This
+    /// lifetime may outlive `self`. If a longer lifetime is not required, or
+    /// you're unsure if you need a longer lifetime, use [value](#method.value).
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let cookie_string = format!("{}={}", "foo", "bar");
+    ///
+    /// // `c` will be dropped at the end of the scope, but `value` will live on
+    /// let value = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.value_raw()
+    /// };
+    ///
+    /// assert_eq!(value, Some("bar"));
+    /// ```
+    #[inline]
+    pub fn value_raw(&self) -> Option<&'c str> {
+        self.cookie_string.as_ref()
+            .and_then(|s| self.value.to_raw_str(s))
+    }
+
+    /// Returns the `Path` of `self` as a string slice of the raw string `self`
+    /// was originally parsed from. If `self` was not originally parsed from a
+    /// raw string, or if `self` doesn't contain a `Path`, or if the `Path` has
+    /// changed since parsing, returns `None`.
+    ///
+    /// This method differs from [path](#method.path) in that it returns a
+    /// string with the same lifetime as the originally parsed string. This
+    /// lifetime may outlive `self`. If a longer lifetime is not required, or
+    /// you're unsure if you need a longer lifetime, use [path](#method.path).
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let cookie_string = format!("{}={}; Path=/", "foo", "bar");
+    ///
+    /// // `c` will be dropped at the end of the scope, but `path` will live on
+    /// let path = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.path_raw()
+    /// };
+    ///
+    /// assert_eq!(path, Some("/"));
+    /// ```
+    #[inline]
+    pub fn path_raw(&self) -> Option<&'c str> {
+        match (self.path.as_ref(), self.cookie_string.as_ref()) {
+            (Some(path), Some(string)) => path.to_raw_str(string),
+            _ => None,
+        }
+    }
+
+    /// Returns the `Domain` of `self` as a string slice of the raw string
+    /// `self` was originally parsed from. If `self` was not originally parsed
+    /// from a raw string, or if `self` doesn't contain a `Domain`, or if the
+    /// `Domain` has changed since parsing, returns `None`.
+    ///
+    /// This method differs from [domain](#method.domain) in that it returns a
+    /// string with the same lifetime as the originally parsed string. This
+    /// lifetime may outlive `self` struct. If a longer lifetime is not
+    /// required, or you're unsure if you need a longer lifetime, use
+    /// [domain](#method.domain).
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let cookie_string = format!("{}={}; Domain=crates.io", "foo", "bar");
+    ///
+    /// //`c` will be dropped at the end of the scope, but `domain` will live on
+    /// let domain = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.domain_raw()
+    /// };
+    ///
+    /// assert_eq!(domain, Some("crates.io"));
+    /// ```
+    #[inline]
+    pub fn domain_raw(&self) -> Option<&'c str> {
+        match (self.domain.as_ref(), self.cookie_string.as_ref()) {
+            (Some(domain), Some(string)) => domain.to_raw_str(string),
+            _ => None,
+        }
+    }
+
 }
 
 /// Wrapper around `Cookie` whose `Display` implementation percent-encodes the
@@ -756,6 +888,52 @@ mod tests {
             .expires(expires).finish();
         assert_eq!(&cookie.to_string(),
                    "foo=bar; Expires=Wed, 21 Oct 2015 07:28:00 GMT");
+    }
+
+    #[test]
+    fn cookie_string_long_lifetimes() {
+        let cookie_string = "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io".to_owned();
+        let (name, value, path, domain) = {
+            // Create a cookie passing a slice
+            let c = Cookie::parse(cookie_string.as_str()).unwrap();
+            (c.name_raw(), c.value_raw(), c.path_raw(), c.domain_raw())
+        };
+
+        assert_eq!(name, Some("bar"));
+        assert_eq!(value, Some("baz"));
+        assert_eq!(path, Some("/subdir"));
+        assert_eq!(domain, Some("crates.io"));
+    }
+
+    #[test]
+    fn owned_cookie_string() {
+        let cookie_string = "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io".to_owned();
+        let (name, value, path, domain) = {
+            // Create a cookie passing an owned string
+            let c = Cookie::parse(cookie_string).unwrap();
+            (c.name_raw(), c.value_raw(), c.path_raw(), c.domain_raw())
+        };
+
+        assert_eq!(name, None);
+        assert_eq!(value, None);
+        assert_eq!(path, None);
+        assert_eq!(domain, None);
+    }
+
+    #[test]
+    fn owned_cookie_struct() {
+        let cookie_string = "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io";
+        let (name, value, path, domain) = {
+            // Create an owned cookie
+            let c = Cookie::parse(cookie_string).unwrap().into_owned();
+
+            (c.name_raw(), c.value_raw(), c.path_raw(), c.domain_raw())
+        };
+
+        assert_eq!(name, None);
+        assert_eq!(value, None);
+        assert_eq!(path, None);
+        assert_eq!(domain, None);
     }
 
     #[test]


SergioBenitez:
I don't think we should go this route. The idea that a simple function can `panic!` because the user doesn't know the internal layout is very unfriendly and likely error-prone. A simpler solution would be to simply expose the indexes into the string so that the user (you) can index the backing string directly; how do you feel about that?

golddranks:
That's not what happens though? It returns an `Option<&str>` which is `Some` or `None` depending on if indexed strings are available.

The panic there was already there before I touched the code. I believe it's just an debug assert, to ensure that the value of CookieStr::Indexed and the existence of the string are in sync.

SergioBenitez:
I wrote all of the `Cookie` code, actually. ;)

I'm talking about the `panic!` in `to_backing_str`.

alexcrichton:
@golddranks don't these methods want to return `Option<&'c str>`, not `Option<&str>`? (the latter being tied to `&self`)

@SergioBenitez I don't think any of these functions panic though? The intention at least is to have an optional method to get a longer lifetime for optimizations (if possible) and then fall back to a shorter lifetime which is guaranteed to succeed if needed.

golddranks:
@alexcrichton Yes, sorry, I've been pushing commits quite haphazardly :D. Sent an new one. Well anyway, what do you think of the general gist? If you have better suggestions for names, I'm all ears.

SergioBenitez:
@alexcrichton Right, I misread the code since the function was changed.

To make my point clear, I don't think we should add another three or four methods to `Cookie` when we could simply expose the underlying indexes to get the same (and slightly more) functionality.

golddranks:
How would you expose the indices? Would you publish the fields...? I'm a bit worried about the ergonomics of that kind of a solution.

SergioBenitez:
Yeah, sure. You'd then be able to use the `to_str` method directly like this:

```rust
if let Some(path) = cookie.path.map(|c| c.to_str(Some(original_str))) {
    // the lifetime of `path` is that of `original_str`
}
```

We could change the signature of `.to_str` to take a `T: Into<Option<&str>>` instead of the `Option` directly so that you don't have to use `Some`:

```rust
if let Some(path) = cookie.path.map(|c| c.to_str(original_str)) {
    // the lifetime of `path` is that of `original_str`
}
```

Is this as nice as `cookie.path_backing()`? No, but 1) it's more flexible since it returns a string even if the underlying string has been changed, and 2) it requires no additional lines of code, and in particular, no methods that may be confusing to users.

alexcrichton:
@SergioBenitez hm I'm curious, how much more flexibility do indices buy us? I'd imagine that returning a `&'c str` would be maximally flexible?

alexcrichton:
(I do agree that the naming here is a bit odd, though)

golddranks:
Lifetime-wise that would be as flexible, but I guess that getting indices would be more flexible in the case that you don't want to get string slices that are "cut" at the predetermined field boundaries. For example, with `&'c str`, you can get the value of the "domain" field: `".example.com"`, but if you want to get the key and the value: `"Domain=.example.com"`, the predetermined API won't give you that.

However, I'd imagine that getting `&'c str` spans 99% of the cases one would use this...

golddranks:
The Regex crate used to return pair of indices for some operations. That felt a bit unergonomic, and now with the `0.2.*` versions it returns a struct that's easy to get the indices from, but it's also supports `.to_str()`. I haven't used that too much, but I think that situation is roughly analogous to this.

SergioBenitez:
@alexcrichton The only additional flexibility you get is that you're able to get the indices (indexes, blasted!) even if the string is `Owned`.

alexcrichton:
@SergioBenitez hm so I personally feel like these are the correct APIs, and I currently disagree that indexes are more flexible. If you've got the original storage on hand you can always derive the indices, and otherwise it seems like 99% of the time you'll just immediately want the actual value anyway, right?

golddranks:
I agree that getting the indices is slightly more flexible, but only very marginally so – and I think that the simplicity and the ergonomics of the API proposed here is enough to outweigh that in comparison. But I still think the names could be better, so if we go with this, let's think about the method names a bit more.

alexcrichton:
Perhaps `raw`? That could emphasize that processed values don't show up (return `None`) but the raw version (original passed in) may come out.

golddranks:
Tests are still not implemented.

golddranks:
Okay, what do you think?

alexcrichton:
@SergioBenitez thoughts on recent comments/developments?

SergioBenitez:
My main concern is that this adds a bunch of new methods to the `Cookie` struct which will rarely be used. At a casual glance, they appear to be duplicating existing effort; this can easily lead to confusion about why the methods exist. I'd like to avoid that.

@alexcrichton The API I propose is both more flexible and less error prone because you can retrieve a string even when the originally parsed value has changed. Consider the following code, as illustration:

```rust
// parse a cookie from an `original_str` and change the value to `"baz"`
let original_str = "foo=bar";
let mut cookie = Cookie::parse(original_str).unwrap();
cookie.set_value("baz");

// this is the API I propose, which mostly just exposes the current internal API
let value = cookie.value.to_str(original_str); // `value` == "baz"

// this is the API proposed in this PR
let value = cookie.value_raw(); // `value` == None
```

Notice that with the API I propose, we get the desired value with the design lifetime, always, as long as it exists. With the API proposed in this PR, we get `None`, even though there's obviously a value.

For these reasons, I am not in favor of accepting this PR. I do believe we should expose a mechanism to retrieve values with a lifetime dictated by the user, and I believe the one I propose is the route we should take.

golddranks:
Note that if you expose just the internal API, that still needs to have two separate methods (similar to `to_str` and `to_raw_str` of `CookieStr` in this branch.) This is because `to_str` must return a str with the lifetime of `Cookie`, because `CookieStr` might have the variant `Concrete` instead of `Indexed`.

To be able to return the lifetime of `original_str`, a separate method like `to_raw_str` is needed.

alexcrichton:
@SergioBenitez such an API seems like it's easy to panic though? If you pass in the wrong "original string" sems like you'd have to panic.

I also feel like exposing literal struct fields is even *more* of an API surface area because it's a whole new type with trait impls to stabilize and such.

And finally, if you just want the value then you call `.value()`, these methods are only intended for when you want to optimize based on the longer lifetime (presumably). There's always of course an infallible method to acquire the vaiours fields.

(also it seems odd to require you to pass in a parameter that the cookie already has stored in itself)

SergioBenitez:
These are all valid points, but I feel strongly that we not add several methods that will rarely be used and can lead to confusion. Is there a better approach? What if we place these methods behind a `RawCookieStr` trait and have `Cookie` implement them? Or maybe put them behind a feature flag?

golddranks:
A feature flag seems like an overkill, or more like just wrong tool for this to me, especially because it's often associated with more configuration, linking and dependency-related things, not just hiding "advanced" functionality.

I think the APIs implemented in this branch are conceptually actually quite simple, and having good docstrings (if you think the current ones are confusing, I can try and improve them) helps to prevent confusion.

Also, I think that an additional trait that is used only for "hiding" methods is an antipattern. I think it would suffice for them to at the low position of the API docs. (Apparently rustdoc preserves the source order.)

alexcrichton:
Yeah in general I'd not advocate trying to hide "advanced" functionality, it just inadvertently punishes those who actually need the behavior.

@SergioBenitez could you list concretely what you see the downsides of adding these methods are? What confusing do you think will arise? Do you think documentation could head this off?

golddranks:
I'd like to depend on this functionality in a code I have to ship soon, so I'd appreciate if we could reach some kind of a agreement soonish. Not urgent, but I'd appreciate it. After all, there is always the option to `.clone()` :P)

alexcrichton:
(ping @SergioBenitez)

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
 
         Ok(())
     }
+
+    /// Returns an `Option` that either contains the name of `self` as a string
+    /// slice of the raw string `self` is originally parsed from, or `None` if
```
SergioBenitez: Let's make this a bit more readable. Let's change this to:

> Returns the name of `self` as a string slice of the raw string `self` was originally parsed from. If `self` was not originally parsed from a raw string, returns `None`.
>
> This method differs from [name](#method.name) in that it returns a string with the same lifetime as the originally parsed string. This lifetime may outlive the `Cookie` struct. If a longer lifetime is not required, or you're ensure if you need a longer lifetime, use [name](#method.name).

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    /// }; // `Cookie` will be dropped here, but the name will live on.
+    /// assert_eq!(name, Some("foo"));
+    /// ```
+    #[inline]
+    pub fn name_raw(&self) -> Option<&'c str> {
+        self.cookie_string.as_ref()
+            .and_then(|s| self.name.to_raw_str(s))
+    }
+
+    /// Returns an `Option` that either contains the value of `self` as a string
```
SergioBenitez: Use the same docstring as the one I suggest for `name_raw` above.

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    /// assert_eq!(value, Some("bar"));
+    /// ```
+    #[inline]
+    pub fn value_raw(&self) -> Option<&'c str> {
+        self.cookie_string.as_ref()
+            .and_then(|s| self.value.to_raw_str(s))
+    }
+
+    /// Returns an `Option` that either contains the `Path` of the cookie as a
+    /// string slice of the raw string `self` is originally parsed from,
```
SergioBenitez: Let's change this to:

> Returns the `Path` of `self` as a string slice of the raw string `self` was originally parsed from. If `self` was not originally parsed from a raw string, or if `self` doesn't contain a `Path`, or if the `Path` has changed since parsing, returns `None`.
>
> This method differs from [path](#method.path) in that it returns a string with the same lifetime as the originally parsed string. This lifetime may outlive the `Cookie` struct. If a longer lifetime is not required, or you're ensure if you need a longer lifetime, use [path](#method.path).

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    ///
+    /// let cookie_string = format!("{}={}; Domain=crates.io", "foo", "bar");
+    /// let domain = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.domain_raw()
+    /// }; // `Cookie` will be dropped here, but the domain will live on.
+    /// assert_eq!(domain, Some("crates.io"));
+    /// ```
+    #[inline]
+    pub fn domain_raw(&self) -> Option<&'c str> {
```
SergioBenitez: Use the same text as what I suggest for `path_raw` above.

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    /// let cookie_string = format!("{}={}; Path=/", "foo", "bar");
+    /// let path = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.path_raw()
+    /// }; // `Cookie` will be dropped here, but the path will live on.
+    /// assert_eq!(path, Some("/"));
+    /// ```
+    #[inline]
+    pub fn path_raw(&self) -> Option<&'c str> {
+        match (&self.path, &self.cookie_string) {
```
SergioBenitez: You shouldn't need the `&`'s here.

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    /// let cookie_string = format!("{}={}; Domain=crates.io", "foo", "bar");
+    /// let domain = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.domain_raw()
+    /// }; // `Cookie` will be dropped here, but the domain will live on.
+    /// assert_eq!(domain, Some("crates.io"));
+    /// ```
+    #[inline]
+    pub fn domain_raw(&self) -> Option<&'c str> {
+        match (&self.domain, &self.cookie_string) {
```
SergioBenitez: You shouldn't need the `&`'s here.

src/lib.rs
```diff
@@ -759,6 +867,70 @@ mod tests {
     }
 
     #[test]
+    fn cookie_string_long_lifetimes() {
+        let cookie_string =
+            "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io".to_owned();
```
SergioBenitez: This should be on one line.

src/lib.rs
```diff
@@ -759,6 +867,70 @@ mod tests {
...
+        let name;
+        let value;
+        let path;
+        let domain;
+        {
+            // Pass a slice:
+            let c = Cookie::parse(cookie_string.as_str()).unwrap();
```
SergioBenitez: Use `&cookie_string`.

src/lib.rs
```diff
@@ -759,6 +867,70 @@ mod tests {
...
 
     #[test]
+    fn cookie_string_long_lifetimes() {
+        let cookie_string =
+            "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io".to_owned();
+        let name;
+        let value;
+        let path;
+        let domain;
+        {
```
SergioBenitez: Can we do `let (name, value, path, domain) = { ..., (c.name_raw(), c.value_raw(), c.path_raw(), c.domain_raw()) }`?

src/lib.rs
```diff
@@ -759,6 +867,70 @@ mod tests {
...
+            // Pass a slice:
+            let c = Cookie::parse(cookie_string.as_str()).unwrap();
+            name = c.name_raw();
+            value = c.value_raw();
+            path = c.path_raw();
+            domain = c.domain_raw();
+        }
```
SergioBenitez: Empty line after this.

src/lib.rs
```diff
@@ -759,6 +867,70 @@ mod tests {
...
+        assert_eq!(name, Some("bar"));
+        assert_eq!(value, Some("baz"));
+        assert_eq!(path, Some("/subdir"));
+        assert_eq!(domain, Some("crates.io"));
+    }
+
+    #[test]
+    fn owned_cookie_string() {
```
SergioBenitez: Same comments as the previous test.

src/lib.rs
```diff
@@ -759,6 +867,70 @@ mod tests {
...
+            domain = c.domain_raw();
+        }
+        assert_eq!(name, None);
+        assert_eq!(value, None);
+        assert_eq!(path, None);
+        assert_eq!(domain, None);
+    }
+
+    #[test]
+    fn owned_cookie_struct() {
```
SergioBenitez: Same here.

src/lib.rs
```diff
@@ -105,13 +98,25 @@ impl CookieStr {
     ///
     /// Panics if `self` is an indexed string and `string` is None.
     fn to_str<'s>(&'s self, string: Option<&'s Cow<str>>) -> &'s str {
-        if self.is_indexed() && string.is_none() {
-            panic!("Cannot convert indexed str to str without base string!")
+        match *self {
```
SergioBenitez: I much prefer the way this was written before. Not only does it look better, the invariant is incredibly clear: `self.is_indexed() => string.is_some()`.

src/lib.rs
```diff
@@ -105,13 +98,25 @@ impl CookieStr {
...
+            CookieStr::Concrete(ref cstr) => &*cstr,
         }
+    }
 
+    fn to_raw_str<'s, 'c: 's>(&'s self, string: &'s Cow<'c, str>) -> Option<&'c str> {
         match *self {
-            CookieStr::Indexed(i, j) => &string.unwrap()[i..j],
-            CookieStr::Concrete(ref cstr) => &*cstr,
+            CookieStr::Indexed(i, j) => {
+                match string {
```
SergioBenitez: `match *string`.

src/lib.rs
```diff
@@ -330,7 +335,7 @@ impl<'c> Cookie<'c> {
     pub fn value(&self) -> &str {
         self.value.to_str(self.cookie_string.as_ref())
     }
-
+
```
SergioBenitez: Trailing whitespace.

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let cookie_string = format!("{}={}", "foo", "bar");
+    /// let name = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.name_raw()
+    /// }; // `Cookie` will be dropped here, but the name will live on.
```
SergioBenitez: Move this comment to the top of the `let` declaration. Add an empty line after this.

src/lib.rs
```diff
@@ -640,6 +645,109 @@ impl<'c> Cookie<'c> {
...
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::Cookie;
+    ///
+    /// let cookie_string = format!("{}={}", "foo", "bar");
+    /// let value = {
+    ///     let c = Cookie::parse(cookie_string.as_str()).unwrap();
+    ///     c.value_raw()
+    /// }; // `Cookie` will be dropped here, but the value will live on.
```
SergioBenitez: Same here, and in the rest of them.

golddranks:
Thanks for the review!

src/lib.rs
golddranks: Can't use that here, because `From<&String>` is not implemented for `Cow<'_, str>`, so I must manually convert to string slice. (Note: Maybe that would be a reasonable impl to have! Ping @alexcrichton )

src/lib.rs
golddranks: If I remove them, it complains that I can't move out from a borrowed context. I think that accessing the fields counts as deref, and because it first constructs a tuple from them instead of matching directly, I need to reference them to prevent moving.

src/lib.rs
golddranks: Same as below.

src/lib.rs
golddranks: I disagree about the clarity; I find the new way clearer. I think that `expect` signals nicely that this is an expected invariant. If this way is used, at least we should use `assert` instead of plain `panic` for signalling intent.

I also dislike "matching twice".

I think that matching once for "indexed" and then if unwrapping with `expect` signals very clearly that string is expected to be some if `CookieStr` is indexed.

golddranks:
Left comments where I couldn't incorporate the changes verbatim.

src/lib.rs
SergioBenitez: Okay, then let's write this (and the other one) as:

```rust
match (self.path.as_ref(), self.cookie_string.as_ref()) {
    (Some(path), Some(string)) => path.to_raw_str(string),
    _ => None
}
```

src/lib.rs
```diff
@@ -640,6 +645,133 @@ impl<'c> Cookie<'c> {
...
         Ok(())
     }
+
+    /// Returns the name of `self` as a string slice of the raw string `self` is
+    /// originally parsed from. If `self` was not originally parsed from a raw
+    /// string, returns `None`.
+    /// 
+    /// This method differs from [name](#method.name) in that it returns a string
+    /// with the same lifetime as the originally parsed string. This lifetime may
+    /// outlive the `Cookie` struct. If a longer lifetime is not required, or
```
SergioBenitez: Actually, we should say (here and in all other new docstrings) "may outlive `self`".

src/lib.rs
```diff
@@ -640,6 +645,133 @@ impl<'c> Cookie<'c> {
...
+    /// 
+    /// assert_eq!(value, Some("bar"));
+    /// ```
+    #[inline]
+    pub fn value_raw(&self) -> Option<&'c str> {
+        self.cookie_string.as_ref()
+            .and_then(|s| self.value.to_raw_str(s))
+    }
+
+    /// Returns the `Path` of `self` as a string slice of the raw string self was
```
SergioBenitez: Use `self` (with ticks) everywhere.

src/lib.rs
```diff
@@ -640,6 +645,133 @@ impl<'c> Cookie<'c> {
...
+    #[inline]
+    pub fn path_raw(&self) -> Option<&'c str> {
+        match (&self.path, &self.cookie_string) {
+            (&Some(ref c), &Some(ref s)) => c.to_raw_str(s),
+            _ => None,
+        }
+    }
+
+    /// Returns the `Domain` of `self` as a string slice of the raw string self was
+    /// originally parsed from. If self was not originally parsed from a raw
```
SergioBenitez: Same with `self` here.

src/lib.rs
```diff
@@ -759,6 +891,55 @@ mod tests {
...
+    }
+
+    #[test]
+    fn owned_cookie_struct() {
+        let cookie_string = "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io";
+
+        let (name, value, path, domain) = {
+            // Create an owned cookie
+            let c = Cookie::parse(cookie_string).unwrap().into_owned();
+
```
SergioBenitez: Kill this empty line.

src/lib.rs
```diff
@@ -759,6 +891,55 @@ mod tests {
...
+        assert_eq!(name, None);
+        assert_eq!(value, None);
+        assert_eq!(path, None);
+        assert_eq!(domain, None);
+    }
+
+    #[test]
+    fn owned_cookie_struct() {
+        let cookie_string = "bar=baz; Path=/subdir; HttpOnly; Domain=crates.io";
+
```
SergioBenitez: I'd kill this empty line in this test and others.

src/lib.rs
golddranks: Nice! I'll do that.

golddranks:
Do you mean lines like `    /// ` by trailing whitespace? I left the whitespace for it to align with the left margin of the text, should I remove those?

src/lib.rs
```diff
@@ -728,25 +728,25 @@ impl<'c> Cookie<'c> {
...
-    /// 
+    ///
     /// assert_eq!(path, Some("/"));
     /// ```
     #[inline]
     pub fn path_raw(&self) -> Option<&'c str> {
-        match (&self.path, &self.cookie_string) {
-            (&Some(ref c), &Some(ref s)) => c.to_raw_str(s),
+        match (self.path.as_ref(), self.cookie_string.as_ref()) {
+            (Some(c), Some(s)) => c.to_raw_str(s),
```
SergioBenitez: Please give the variables descriptive names. `path`, `string` work.

src/lib.rs
```diff
@@ -755,19 +755,19 @@ impl<'c> Cookie<'c> {
...
-    /// 
+    ///
     /// assert_eq!(domain, Some("crates.io"));
     /// ```
     #[inline]
     pub fn domain_raw(&self) -> Option<&'c str> {
-        match (&self.domain, &self.cookie_string) {
-            (&Some(ref c), &Some(ref s)) => c.to_raw_str(s),
+        match (self.domain.as_ref(), self.cookie_string.as_ref()) {
+            (Some(c), Some(s)) => c.to_raw_str(s),
```
SergioBenitez: Same here.

src/lib.rs
```diff
@@ -89,13 +89,6 @@ enum CookieStr {
 }
 
 impl CookieStr {
-    /// Whether this string is derived from indexes or not.
-    fn is_indexed(&self) -> bool {
-        match *self {
-            CookieStr::Indexed(..) => true,
-            CookieStr::Concrete(..) => false,
-        }
-    }
```
SergioBenitez: Kill this empty line.

SergioBenitez:
Your last commit got rid of the trailing whitespace, so that's fine.

golddranks:
Just a note in case the latest commits didn't notify this already: I've addressed all the comments.

src/lib.rs
```diff
@@ -640,6 +644,133 @@ impl<'c> Cookie<'c> {
 
         Ok(())
     }
+
+    /// Returns the name of `self` as a string slice of the raw string `self` is
```
SergioBenitez: "is" should be "was", here and elsewhere.

src/lib.rs
```diff
@@ -640,6 +644,133 @@ impl<'c> Cookie<'c> {
...
+    /// originally parsed from. If `self` was not originally parsed from a raw
+    /// string, returns `None`.
+    ///
+    /// This method differs from [name](#method.name) in that it returns a string
+    /// with the same lifetime as the originally parsed string. This lifetime may
+    /// outlive `self`. If a longer lifetime is not required, or
```
SergioBenitez: Word wrap is incorrect here and in other places.

golddranks: Which width should I use? I noticed that most of the original code wrapped around 80 but didn't follow this strictly.

SergioBenitez: 80 cols for docstrings and comments, 100 for regular code; this is the standard.

golddranks: Hm, the comments seem always to wrap at 80, even though some spurious code lines go further. Is this how it should go?

golddranks:
Is Travis stuck or does it usually take many hours to get started?

golddranks:
The checks passed. @alexcrichton and @SergioBenitez, if you don't have any further comments, I'd appreciate if this could be pulled in (and preferable released as a a minor version) today :)

alexcrichton:
Thanks @golddranks!

golddranks:
Thanks you!

PR #79, opened 2017-02-22
Restore rustc-serialize functionality

nox:


nox:
Blocked by https://github.com/rust-lang-deprecated/time/pull/144. Needed to bump Hyper in Servo.

alexcrichton:
@nox did you mean to add serde support instead of rustc-serialize? That's what https://github.com/servo/servo/issues/15010 seems to indicate at least.

Also, can you back out the version bump? There's a few other changes in flight and I'd like to take that one step at a time.

finally, cc @SergioBenitez

nox:
No I meant `rustc-serialize` because it was what the Servo code previously used but you broke support for it without telling us. So instead of waiting for this couple of PRs to land, I switched `net` in Servo to `serde_json` for cookie persistance.

nox:
I removed the version bump from this PR.

alexcrichton:
@nox I'm not quite understanding, you mention that you're switching to serde but require rustc-serialize in this PR? Adding this support is fine, it just seems odd...

In any case could the serialized version use the stringified value of the cookie rather than the in-memory representation? That should be much more stable across versions of cookie and not require changes to the time crate as well.

nox:
> @nox I'm not quite understanding, you mention that you're switching to serde but require rustc-serialize in this PR? Adding this support is fine, it just seems odd...

The cookie crate still has a `rustc-serialize` feature that just does nothing, that's why I made a couple of PRs to restore that `rustc-serialize` support, because it being missing caused some issues in Servo.

Meanwhile, and *after* I filed this couple of PRs, I took this opportunity to just switch to `serde_json` for the things that required `rustc-serialize` support in `cookie`, that's all.

nox:
> In any case could the serialized version use the stringified value of the cookie rather than the in-memory representation? That should be much more stable across versions of cookie and not require changes to the time crate as well.

I only *restored* the previous `rustc-serialize` support that was broken in #70, and given I don't need this PR anymore, I don't want to lose any more time on this.

alexcrichton:
@nox so... this PR isn't needed any more? Sorry I'm still confused a bit as to what's going on?

nox:
I did it anyway because there is still a [`rustc-serialize` feature](https://github.com/alexcrichton/cookie-rs/blob/master/Cargo.toml#L23) in that crate, so I thought the removal of it was unintentional.

alexcrichton:
Ah yes I believe leaving that in was a mistake. In the recent refactoring it was decided to remove this feature and add it back if necessary. 

So is this PR no longer needed for servo?

nox:
No.

PR #85, opened 2017-05-07
Switch from time to chrono

erickt:
This is a breaking change. The time module is deprecated, and suggests switching over to the chrono library, which is more actively maintained. One thing to note is that chrono does not support parsing with `%Z`:

https://github.com/chronotope/chrono/issues/38

However, this actually isn't that bad. According to RFC 2616, the only supported timezone is "GMT":

http://tools.ietf.org/html/rfc2616#section-3.3.1

Furthermore, @lifthrasiir points out that most implementations of `%Z` is actually not widely supported. On my mac, it claims to only support parsing the local timezone, and GMT, so I believe it's safe to hardcode the timezone.

This also fixes a bug in an example, where October 21, 2017 would be a Saturday. However, Wednesday 21 October, 2015 is Back to the Future day, so this switches to using that date instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ secure = ["ring", "base64"]
 percent-encode = ["url"]
 
 [dependencies]
-time = "0.1"
+chrono = "0.3.1"
 url = { version = "1.0", optional = true }
 ring = { version = "0.7.4", optional = true }
 base64 = { version = "0.4", optional = true }
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 
-use time::{Tm, Duration};
+use chrono::{DateTime, Duration, UTC};
 
 use ::{Cookie, SameSite};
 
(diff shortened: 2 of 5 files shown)


erickt:
True :) I did try to ping you yesterday to chat about time vs chrono, but I figured it'd be fun to start poking at this since `time` is deprecated.

@lifthrasiir did add a comment in the [readme](https://github.com/chronotope/chrono/blob/master/README.md#duration) that the intention is to switch to their own implementation of `Duration`, which I think would fully decouple them from the `time` crate. So we could switch to `chrono`, and just transparently pick up the new `Duration` once it happens. Or I suppose another option is that according to the [spec](https://tools.ietf.org/html/rfc6265#section-5.2.2), it is supposed to just be seconds, so we could just parse into a u64?

alexcrichton:
I'd be fine just going into a u64 for now and relying on helper methods elsewhere to convert to chrono/std/whatever. @SergioBenitez may wish to weigh in before merging though.

nokaa:
One reason to switch to chrono is to make it easier for dependent libraries. Currently I have an application that uses chrono for postgres, but also requires time for cookies. It would be nice to only use chrono.

SergioBenitez:
Ideally we wouldn't use either `chrono` or `time`. I'd rather stick with `time` than move to `chrono`, however, not least because it doesn't buy us anything at the moment.

@alexcrichton I'd be in favor of using `u64`, though it does make things less convenient. We could, alternatively, reexport the appropriate `time` types.

SergioBenitez:
Looks like there's consensus on not moving to `chrono`. As a result, I'm going to close this out.

karuna:
@SergioBenitez is it possible to reconsider this? Since rocket_db_pools supports sqlx and sqlx have chrono as the feature-flag.

PR #86, opened 2017-05-07
Don't panic if passed in a large max age.

erickt:
Duration will panic if the number of seconds is greater than 2^63/1000. This just caps the Max-Age to the highest value we can parse.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::cmp;
 use std::error::Error;
 use std::ascii::AsciiExt;
 use std::str::Utf8Error;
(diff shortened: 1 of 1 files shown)


src/parse.rs
```diff
@@ -152,7 +152,16 @@ fn parse_inner<'c>(s: &str, decode: bool) -> Result<Cookie<'c>, ParseError> {
                 // max age as 0 seconds.
                 cookie.max_age = match v.parse() {
                     Ok(val) if val <= 0 => Some(Duration::zero()),
-                    Ok(val) => Some(Duration::seconds(val)),
+                    Ok(val) => {
+                        // Don't panic if the max age seconds is greater than what's supported by
+                        // `Duration`.
+                        let max_value = Duration::max_value().num_seconds();
```
SergioBenitez: Can we use `min` here instead? Something like:

```rust
let capped_val = cmp::min(Duration::max_value().num_seconds(), val);
Some(Duration::seconds(capped_val))
```

SergioBenitez:
This is a great catch. Thanks for this!

PR #88, opened 2017-05-08
0.8.0: Update to base64 0.5.2 and *ring* 0.9.1.

briansmith:
base64 0.5.2 fixes a security vulnerability. *ring* 0.9 is used by the
latest version of Rustls and some other crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 
 name = "cookie"
 authors = ["Alex Crichton <alex@alexcrichton.com>", "Sergio Benitez <sb@sergio.bz>"]
-version = "0.7.6"
+version = "0.8.0"
 license = "MIT/Apache-2.0"
 repository = "https://github.com/alexcrichton/cookie-rs"
 documentation = "https://docs.rs/cookie"
@@ -18,5 +18,5 @@ percent-encode = ["url"]
 [dependencies]
 time = "0.1"
 url = { version = "1.0", optional = true }
-ring = { version = "0.7.4", optional = true }
-base64 = { version = "0.4", optional = true }
+ring = { version = "0.9.1", optional = true }
+base64 = { version = "0.5.2", optional = true }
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks!

alexcrichton:
Er wait oops sorry, didn't realized that the version bump was in here! @briansmith AFAIK ring/base64 are private deps right now, so I think this is a non-breaking change, right?

briansmith:
I'm not sure if it is considered a breaking change or not. With this change, cookie-rs will no longer be compatible with anything that was using versions of *ring* before 0.9. If you don't consider that a breaking change then I am happy to undo the version bump.

alexcrichton:
Oh if that's the case I'm totally fine bumping the version of cookie, but could you also elaborate on the breakage? For example if I've got a crate `foo` which depends on *ring* 0.8 and the master branch of this repo (cookie with a dep on *ring* 0.9), will that not compile? (will it get a link error?)

briansmith:
Somebody reported that when *ring* 0.7 and *ring* 0.9 are linked together, the linker will link in the C/asm code from 0.9 and use it to satisfy the symbols referenced from *ring* 0.7, which causes things to go very wrong. (I have no clue how to prevent the toolchain from linking a program that includes two different versions of *ring*.)

alexcrichton:
Ah ok, thanks for the explanation! I'll leave a comment over at https://github.com/briansmith/ring/issues/535 to see if I can help out.

In any case, thanks so much for the PR!

alexcrichton:
@SergioBenitez I was a bit too fast to pull the trigger on this one I think. Just a heads up that upgrading to *ring* 0.9 will require a major version bump of the cookie crate as well. It seems fine by me to happen whenever, but if you want to delay the 0.7 -> 0.8 update we may want to revert the merge commit temporarily, re-merge the base64 update, and then have a pending PR for *ring* 0.9 at least for when we're ready for 0.8

PR #91, opened 2017-05-19
Add Cookie methods that can clear same_site/max_age/path/domain

erickt:
This allows us to remove the values for Option types.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -679,6 +679,85 @@ impl<'c> Cookie<'c> {
         self.expires = Some(time);
     }
 
+    /// Clears the value of `same_site` in `self`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use cookie::{Cookie, SameSite};
+    ///
+    /// let mut c = Cookie::new("name", "value");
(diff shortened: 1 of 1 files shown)


SergioBenitez:
Should we instead take in a `T: Into<Option<_>>` in the existing `set_` functions?

erickt:
Sure, that'd work too. I'll update my PR.

erickt:
@SergioBenitez: I don't think this will quite work. For example:

```rust
    pub fn set_path<T, P>(&mut self, path: T)
        where T: Into<Option<P>>,
              P: Into<Cow<'static, str>>,
    {
        self.path = path.into().map(|path| {
            CookieStr::Concrete(path.into())
        });
    }
```

Rust gets rather confused when you try to do `cookie.set_path(None);`. I still haven't yet figured out what type annotation I need to specify in order to get it to work.

What if we just change the method to take an `Option<_>` type?

Boscop:
Any update on this? :)
We need a way to not set SameSite at all.

SergioBenitez:
@Boscop I've added the `SameSite::None` variant to handle this case.

PR #92, opened 2017-05-19
Switch max-age over to u64

erickt:
This is a breaking change. In #85, it was suggested that we remove the public dependency on `time::Duration` since that crate is deprecated. According to the [Cookie spec](https://tools.ietf.org/html/rfc6265#section-5.2.2), `Max-Age` is in seconds that is either zero or always positive. Negative numbers are always parsed to the earliest representable date and time, which would be 0.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 
-use time::{Tm, Duration};
+use time::Tm;
 
 use ::{Cookie, SameSite};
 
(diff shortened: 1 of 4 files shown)


SergioBenitez:
Can we get rid of all of `time`? It seems inconsistent to use actual structures for some things but not others.

erickt:
What would we use to parse "Expires" if not time or chrono? Leave it unparsed, or should cookie grow its own time structure that uses time or chrono under the covers? That seems a tad sad but I suppose it makes sense until we have a stable time crate.

SergioBenitez:
By "get rid of time" I mean "don't expose `time`". it's fine to have an internal dependency, whichever it might be.

SergioBenitez:
We've moved over to `time` 0.2, so this PR is no longer applicable.

PR #93, opened 2017-05-22
Clamp "Expires" to 9999 years

erickt:
According to [rfc 6265](https://tools.ietf.org/html/rfc6265#section-5.1.1), cookies support at most 9999 years.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -675,7 +675,12 @@ impl<'c> Cookie<'c> {
     /// # }
     /// ```
     #[inline]
-    pub fn set_expires(&mut self, time: Tm) {
+    pub fn set_expires(&mut self, mut time: Tm) {
+        // According to https://tools.ietf.org/html/rfc6265#section-5.1.1, cookie expires needs 
+        // to be less than or equal to 9999 years. Note that Tm.tm_year starts at 1900, so shift
+        // our value back to get the correct time.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -675,7 +675,11 @@ impl<'c> Cookie<'c> {
     /// # }
     /// ```
     #[inline]
-    pub fn set_expires(&mut self, time: Tm) {
+    pub fn set_expires(&mut self, mut time: Tm) {
+        // According to https://tools.ietf.org/html/rfc6265#section-5.1.1, cookie expires needs 
+        // to be less than or equal to 9999 years.
+        if time.tm_year > 8099 { time.tm_year = 8099; }
```
SergioBenitez: Is it `8099` or `9999`? Are you accounting for the possibility that a `Tm` may not make sense, i.e, where `month` might be 2^16 or something nonsensical?

SergioBenitez:
There's a build failure.

src/lib.rs
erickt: I'll add a comment. `Tm.tm_year` starts at 1900, so the max value we can have is 8099 + 1900 = 9999.

SergioBenitez:
Merged in https://github.com/SergioBenitez/cookie-rs/commit/121d6887131e52b38cdf0b39a067c25651f89ed5.

PR #99, opened 2017-10-12
Add cookie method expire now

ernestas-poskus:
cc @SergioBenitez 

Adding a convenience method to expire cookie.


ernestas-poskus:
cc @alexcrichton

SergioBenitez:
I'm not in favor of adding this. It's easy enough to do outside of the crate and I don't see a great reason to add this here.

PR #100, opened 2017-11-25
add add_original to private and signed jars

ivstas:
Hi.
While trying to make a pr for this issue https://github.com/SergioBenitez/Rocket/issues/368 I found out that it requires to add original cookie for a `PrivateJar`.

I realised that `add_original` method is missing for some reason. Do you find it suitable to be added to `PrivateJar` and `SignedJar`?

### DIFF
--- a/src/secure/private.rs
+++ b/src/secure/private.rs
@@ -110,7 +110,16 @@ impl<'a> PrivateJar<'a> {
         self.parent.add(cookie);
     }
 
-    /// Adds an "original" `cookie` to this jar.
+    /// Adds an "original" `cookie` to parent jar.
+    /// The cookie's value is encrypted with authenticated encryption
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Awesome, thanks for the PR! The code looks generally good. The documentation for the new `add_original` methods is a bit lackluster, however. Can you add more information to these methods, perhaps using the wording from the existing `add_original` method? Also, can you add some tests for these new methods? Doctests in the docstring are fine.

SergioBenitez:
Thanks so much! This was merged in a53d0069d43dffbd4862d885270a554702b676ae. :)

P.S: To use the latest version of `cookie` in Rocket, simply change the `rev` for `cookie` in Rocket's `Cargo.toml` to `f4e6930`.

PR #104, opened 2018-06-18
Bump ring to fix build error

BartG95:
alpha3 fixed a bug that caused a build error on windows (https://github.com/briansmith/ring/commit/40963edc4e4b66922529227019a8654e80bd8d25).


SergioBenitez:
Fixed in master.

PR #106, opened 2018-07-14
update ring to 0.13

dgriffen:
ring published 0.13 today. The new version fixes the windows compilation issues that were present in 0.12.1 which means that cookie will be able to build on windows again.


SergioBenitez:
Fixed in master. New release will be pushed imminently.

PR #107, opened 2018-10-11
Add 'CookieJar::parse_from_header' to parse a CookieJar from a HTTP Cookie header

mehcode:
```rust
let jar = CookieJar::parse_from_header("key=value; other=other value")?;
```

---

Been using this method in my projects for a while now and decided to send a PR to see if there is interest in adding it to `cookie`. I'll add tests and improve documentation if its decided that this should be added.

---

In the same vein I also have a `CookieJar::to_header_string()` method I'd like to contribute as well. It takes a `CookieJar` and writes it to a semicolon-delimited key=value string like in the HTTP cookie header.

### DIFF
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -6,7 +6,7 @@ use time::{self, Duration};
 #[cfg(feature = "secure")]
 use secure::{PrivateJar, SignedJar, Key};
 use delta::DeltaCookie;
-use Cookie;
+use {Cookie, ParseError};
 
 /// A collection of cookies that tracks its modifications.
 ///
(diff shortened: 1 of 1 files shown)


CrowdHailer:
This is just what I have been looking for. Is there anything blocking it getting merged?

SergioBenitez:
The functionality is far too subjective, in my opinion, for inclusion in the library. For instance, this implementation assumes _at least_ the following:

1. You want to add the cookies with `add` and not `add_original`.
2. You want to keep the latter of a duplicate if one exists.
3. You want to ignore parsing errors as you go.
4. You want the cookies in a regular `CookieJar`.

What's more, implementing this in your own library or application is straightforward, as evidenced by the simplicity of the function.

PR #110, opened 2019-01-10
bump ring to 0.14.0

ubnt-intrepid:


landesfeind:
Hi, would it be possible to go forward with this request? As far as I can tell, the Travis checks for nightly builds failed due to the inability to install nightly. 
Bumping this version would be very helpful!

ubnt-intrepid:
@alexcrichton could you re-run the CI build?

alexcrichton:
cc  @SergioBenitez

SergioBenitez:
It looks like we're getting a true failure. Can you investigate, @ubnt-intrepid?

ubnt-intrepid:
@SergioBenitez The failure is caused by `Iter` returning removal Cookie (it is reproduced on master branch and hence is not unique problem of this PR). the failure occurs only on nightly toolchain, and hence it might be a bug in the standard library.

Test code (modification of `Iter`'s doctest):

```rust
#[test]
fn issue() {
    let mut jar = CookieJar::new();

    jar.add_original(Cookie::new("name", "value"));
    jar.add_original(Cookie::new("second", "two"));

    jar.add(Cookie::new("new", "third"));
    jar.add(Cookie::new("another", "fourth"));
    jar.add(Cookie::new("yac", "fifth"));

    jar.remove(Cookie::named("name"));
    jar.remove(Cookie::named("another"));

    for cookie in jar.iter() {
        eprintln!("({}, {})", cookie.name(), cookie.value());
    }
}
```

Expected output (the result on stable/beta):

```
(new, third)
(yac, fifth)
(second, two)
```

The result on nightly:

```
(name, value)
(second, two)
(yac, fifth)
(new, third)
```

ubnt-intrepid:
Due to the above patch, there is a possibility that the reference to `DeltaCookie` received in `Iter` will be from (unmodified) `original_cookies`.  As a result, [the validation in `Iter::next`](https://github.com/alexcrichton/cookie-rs/blob/5d6c8174990e20ba8f849b40a1d74a37fa6576e4/src/jar.rs#L455) is always true and `Iter` returns its reference even if it is a removed Cookie.

SergioBenitez:
@ubnt-intrepid Thanks for investigating that! Can you rebase on the latest master to rerun tests?

lucab:
Gentle ping to get this reviewed, merged and released.
(there is a bump for the native library `ring-asm` hiding in here, which makes updates in consumers dependent on having all libraries updated first)

Darkspirit:
It looks like crates.io itself is waiting for this: https://github.com/briansmith/ring/issues/774#issuecomment-457859014

dwijnand:
I agree, cutting a new version of cookie with this would probably help the ecosystem recover from the ring versions purge.

PR #111, opened 2019-01-16
force the order of the internal iterators in Iter

ubnt-intrepid:
This patch prevents the possibility that the order of internal iterators in [`hash_set::Union`](https://doc.rust-lang.org/nightly/std/collections/hash_set/struct.Union.html) will be changed, caused by the change in rust-lang/rust#57043.

### DIFF
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -344,7 +344,10 @@ impl CookieJar {
     /// }
     /// ```
     pub fn iter(&self) -> Iter {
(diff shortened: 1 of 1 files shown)


alexcrichton:
@SergioBenitez discovered [here](https://github.com/rust-lang/rust/issues/57943) it looks like this this actually fixes a bug too, so this may want to be published as a release!

PR #121, opened 2019-07-23
Use the radix64 crate to perform base64 encoding/decoding.

ggriffiniii:
Cards on the table: I'm the author of the radix64 crate.

This is a bit of self promotion and feel free to raise any concerns. The radix64 crate is faster at encoding and decoding both large and small inputs. The MSRV of radix64 is 1.31.0, which appears to match the already existing dependencies with the "secure" feature enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,14 +11,14 @@ and private (encrypted + signed) jars.
 """
 
 [features]
-secure = ["ring", "base64"]
+secure = ["ring", "radix64"]
 percent-encode = ["url"]
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


SergioBenitez:
While `radix64` may be faster, it uses roughly 42 counts of `unsafe` to do so. This represents more than 1% of non-comment/blank code. By contrast, `base64` uses no `unsafe` code. As a result, there is significantly more room for memory unsafety. I'd be happy to consider migrating to `radix64` is the `unsafe` surface area is significantly reduced. For now, I'm closing this PR.

PR #122, opened 2019-07-30
Upgrade to url 2.0 ecosystem

benesch:
The url crate recently released a backwards-incompatible update, v2.0.
As part of this upgrade, the percent-encoding code has been moved into
its own crate, titled percent-encoding, and no longer ships standard
percent encoding sets.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,11 +12,11 @@ and private (encrypted + signed) jars.
 
 [features]
 secure = ["ring", "base64"]
-percent-encode = ["url"]
+percent-encode = ["percent-encoding"]
 
 [dependencies]
 time = "0.1"
-url = { version = "1.0", optional = true }
+percent-encoding = { version = "2.0", optional = true }
 ring = { version = "0.14.0", optional = true }
 base64 = { version = "0.10.0", optional = true }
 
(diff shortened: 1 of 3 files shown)


eoger:
Ping @SergioBenitez mind having a look? Looks like this is blocking `reqwest` from upgrading to `url` 2.0. Thanks!

SergioBenitez:
Hi @benesch! Thanks for the PR.

Is this actually encoding the same set of characters? Based on https://docs.rs/percent-encoding/2.1.0/src/percent_encoding/lib.rs.html#99, the `CONTROLS` set does not encode all characters above `0x7E` (`~`), which is required by the standard for the [C0 control percent encode set](https://url.spec.whatwg.org/#c0-control-percent-encode-set). Instead, it only encodes the controls, which means that `USERINFO_ENCODE_SET` is necessarily wrong. If this is correct, we'll need to address this before the PR can be merged.

SergioBenitez:
> Ping @SergioBenitez mind having a look? Looks like this is blocking `reqwest` from upgrading to `url` 2.0. Thanks!

Why is this blocking an upgrade in a different crate? This is a private, non`-sys` dependency, so it should have no bearing on whether another crate can upgrade their dependencies.

alex:
Is there anything I could do to be helpful in getting a release out? (I'm making my quarterly attempt to remove duplicate crates from my cargo.locks :-))

PR #124, opened 2019-09-09
Update *ring* dependency to 0.16

jebrosen:
Commits broken up by file/purpose for easier auditing.

I've attempted to adapt the code to the new APIs in the most straightforward way I could find. The behavior is intended to be identical and cookies signed or encrypted with cookie `0.12` can still be verified and/or decrypted after this PR and vice versa. However, I have not yet confirmed that that is actually the case.


src/secure/key.rs
```diff
@@ -1,13 +1,11 @@
...
+use secure::ring::hkdf::{HKDF_SHA256, Algorithm, Prk, KeyType};
 use secure::ring::rand::{SecureRandom, SystemRandom};
 
 use secure::private::KEY_LEN as PRIVATE_KEY_LEN;
 use secure::signed::KEY_LEN as SIGNED_KEY_LEN;
 
-static HKDF_DIGEST: &'static Algorithm = &SHA256;
-const KEYS_INFO: &'static str = "COOKIE;SIGNED:HMAC-SHA256;PRIVATE:AEAD-AES-256-GCM";
+static HKDF_DIGEST: Algorithm = HKDF_SHA256;
+const KEYS_INFO: &[&[u8]] = &[b"COOKIE", b"SIGNED:HMAC-SHA256", b"PRIVATE:AEAD-AES-256-GCM"];
```
jebrosen: This changes behavior (lost `;`).

src/secure/private.rs
```diff
@@ -144,32 +144,34 @@ impl<'a> PrivateJar<'a> {
...
+        // Randomly generate the nonce
+        let mut nonce_bytes = [0; NONCE_LEN];
+        SystemRandom::new().fill(&mut nonce_bytes).expect("couldn't random fill nonce");
+        let nonce = Nonce::try_assume_unique_for_key(&nonce_bytes)
+            .expect("invalid length of `nonce`");
+
+        // Use cookie's name as associated data to prevent value swapping.
+        let ad = Aad::from(cookie.name().as_bytes());
+
+        // Perform the actual sealing operation and get the output length.
```
jebrosen: This comment about output length is outdated

SergioBenitez:
> However, I have not yet confirmed that that is actually the case.

Please do ping me as soon as you do verify this.

jebrosen:
I've fixed round-trip and added a few test cases that pass both before and after the upgrade.

Cookies were generated with the following code:

<details>

```rust
use std::io;

use cookie::{Cookie, CookieJar, Key};

fn get_line() -> String {
    let mut input = String::new();
    io::stdin().read_line(&mut input).expect("read input");
    if input.as_bytes().last() == Some(&b'\n') {
        input.pop().unwrap();
    }
    input
}

fn get_from_stdin(reason: &str) -> Cookie<'static> {
    println!("Enter name (for {}):", reason);
    let name = get_line();
    println!("Enter value (for {}):", reason);
    let value = get_line();

    Cookie::new(name, value)
}

fn main() {
    // Secret is 'Super secret!' passed through sha256
    let secret = b"\x1b\x1c\x6f\x1b\x31\x99\x82\x77\x0e\x05\xb6\x05\x54\x0b\xd9\xea\x54\x9f\x9a\x56\xf4\x0f\x97\xdc\x6e\xf2\x89\x86\x91\xe0\xa5\x79";
    let key = Key::from_master(secret);

    let mut jar = CookieJar::new();

    let mut signed = jar.signed(&key);
    signed.add(get_from_stdin("signed"));

    let mut private = jar.private(&key);
    private.add(get_from_stdin("private"));

    for cookie in jar.iter() {
        println!("{} => {}", cookie.name(), cookie.value());
    }
}
```

</details>

SergioBenitez:
Merged in 1c3bc2a84347e48cecc991e928b6a20ef7e416d1. Thank you!

PR #127, opened 2019-11-05
Format code using 'cargo fmt'

Atul9:


SergioBenitez:
Some of these are good, some aren't. In general, I don't believe rustfmt is useable quite yet.

PR #129, opened 2019-11-26
Changed behavior of SameSite=None

LucFauvel:
In accordance with the [working draft of the IETF](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00), SameSite=None must be set in specific cases, more can be read [here](https://web.dev/samesite-cookie-recipes/).

These changes rectify this behavior and allows for SameSite to be set as None as well as not being set at all.

### DIFF
--- a/src/draft.rs
+++ b/src/draft.rs
@@ -21,8 +21,10 @@ pub enum SameSite {
     Strict,
     /// The "Lax" `SameSite` attribute.
     Lax,
-    /// No `SameSite` attribute.
-    None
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Thanks for bringing these changes to my attention! And, thank you for the implementation! After consideration, I opted for a slightly different API than the one in your PR; you can see the changes in ae6d4230cac4d3230ae848c76de0862835ecaa8e. As a result, I'm closing this PR. Nevertheless, this PR is the direct cause of these changes, so thank you!

PR #134, opened 2020-01-16
Upgrade to time 0.2

djc:


src/lib.rs
```diff
@@ -805,11 +806,11 @@ impl<'c> Cookie<'c> {
         }
 
         if let Some(max_age) = self.max_age() {
-            write!(f, "; Max-Age={}", max_age.num_seconds())?;
+            write!(f, "; Max-Age={}", max_age.whole_seconds())?;
         }
 
         if let Some(time) = self.expires() {
-            write!(f, "; Expires={}", time.rfc822())?;
+            write!(f, "; Expires={}", time.format("%a, %d %b %Y %H:%M:%S GMT"))?;
```
jhpratt: Be sure to convert the value to UTC before formatting.

SergioBenitez:
@jhpratt What's the motivation for upgrading to 2018?

jhpratt:
@SergioBenitez I suppose it's not necessary, but can't hurt, given that the MSRV is already greater than when it was introduced. I imagine an increasingly greater percentage of Rust users are only familiar with 2018, which would impact the number who are able to contribute meaningfully. Some differences (like paths) are non-trivial.

Cargo.toml
```diff
@@ -15,7 +15,7 @@ secure = ["ring", "base64"]
 percent-encode = ["percent-encoding"]
 
 [dependencies]
-time = "0.1"
+time = "0.2.2"
```
jhpratt: v0.2.3 has been released. It's probably best to disable default features, which currently only consists of deprecated methods.

src/parse.rs
```diff
@@ -237,10 +252,16 @@ pub fn parse_cookie<'c, S>(cow: S, decode: bool) -> Result<Cookie<'c>, ParseErro
     Ok(cookie)
 }
 
+pub(crate) fn parse_gmt_date(s: &str, format: &str) -> Result<OffsetDateTime, time::ParseError> {
+    let primitive = time::PrimitiveDateTime::parse(s, format)?;
```
jhpratt: Time now has a top-level parse function, allowing for type inference. `time::parse(s, format)?` may suffice. I haven't tested it, though.

src/parse.rs
```diff
@@ -237,10 +252,16 @@ pub fn parse_cookie<'c, S>(cow: S, decode: bool) -> Result<Cookie<'c>, ParseErro
...
+    Ok(primitive.using_offset(time::UtcOffset::UTC))
```
jhpratt: You can now use `offset!(UTC)` if wanted. Zero difference either way.

src/parse.rs
SergioBenitez: This is a concerning comment. How much trust do you have in time v0.2, @jhpratt? Can this library trust it? Is the library ready for use? I don't see that any dependent crates have updated to 0.2 (though I didn't exhaustively check), and the library seems to be missing some important features, like time-zone support. Are we sure we want to do this now and not at some later time?

jhpratt: @SergioBenitez To be clear, when I say "I haven't tested it", I'm referring to my suggestion of using that function and relying on type inference in that specific line. The function itself has definitely been tested and verified to work (all it does is forward the call to the relevant type, based on a sealed trait).

I can't say what crates have upgraded to 0.2, though I'm confident in it. I'm using it extensively in a moderately sized (internal) project.

What sort of time zone support would you like to see, tzdb? The only thing that v0.2 changed from v0.1 in this sense is the removal of `%Z`, which didn't even parse time zones — it only handled "UTC" and "GMT" before to maintain compatibility with C. So that change, by itself, should be a non-issue.

src/lib.rs
djc: Fixed, thanks!

Cargo.toml
djc: Updated.

src/parse.rs
djc: @SergioBenitez FWIW, in my view time 0.2 is not worse than 0.1 and is better in many important ways -- especially in the sense that it is maintained, unlike 0.1, and it has a much more idiomatic API.

src/parse.rs
djc: I don't think the macro adds value here.

src/builder.rs
```diff
@@ -65,19 +65,20 @@ impl<'c> CookieBuilder<'c> {
...
     /// # extern crate cookie;
     /// extern crate time;
+    /// use time::OffsetDateTime;
     ///
     /// use cookie::Cookie;
     ///
     /// # fn main() {
     /// let c = Cookie::build("foo", "bar")
-    ///     .expires(time::now())
+    ///     .expires(OffsetDateTime::now())
```
SergioBenitez: This does something different than `time::now()`, correct? `time::now()` returned a `Tm` for the current time in the current time zone, while `OffsetDateTime::now()` returns the current time in UTC. This is fairly surprising. Not only that, but `OffsetDateTime::now()` is available only when the `alloc` feature isn't enabled. Since features are additive, and thus Cargo will only compile a library once (versions permitting) with the union of all requested features, this means dependencies can make this code fail. Now, this is just a test, but we're providing it as an example, so we should make sure it's exemplary.


This is really a problem in the `time` crate; this is an incorrect use of features.

src/lib.rs
```diff
@@ -783,15 +783,16 @@ impl<'c> Cookie<'c> {
...
     /// use cookie::Cookie;
     ///
     /// # fn main() {
     /// let mut c = Cookie::new("name", "value");
     /// assert_eq!(c.expires(), None);
     ///
-    /// let mut now = time::now();
-    /// now.tm_year += 1;
+    /// let mut now = OffsetDateTime::now();
+    /// now += Duration::weeks(52);
```
SergioBenitez: Very odd that there's no `Duration::years(1)`...

src/lib.rs
```diff
@@ -858,11 +859,12 @@ impl<'c> Cookie<'c> {
...
 
         if let Some(max_age) = self.max_age() {
-            write!(f, "; Max-Age={}", max_age.num_seconds())?;
+            write!(f, "; Max-Age={}", max_age.whole_seconds())?;
         }
 
         if let Some(time) = self.expires() {
-            write!(f, "; Expires={}", time.rfc822())?;
+            let time = time.to_offset(UtcOffset::UTC);
+            write!(f, "; Expires={}", time.format("%a, %d %b %Y %H:%M:%S GMT"))?;
```
SergioBenitez: I think this is confusing; we _always_ convert to UTC and emit the date/time as GMT, regardless of what the user desired. Should we really be doing that?

src/parse.rs
```diff
@@ -207,10 +222,10 @@ fn parse_inner<'c>(s: &str, decode: bool) -> Result<Cookie<'c>, ParseError> {
                 // Try strptime with three date formats according to
                 // http://tools.ietf.org/html/rfc2616#section-3.3.1. Try
                 // additional ones as encountered in the real world.
-                let tm = time::strptime(v, "%a, %d %b %Y %H:%M:%S %Z")
-                    .or_else(|_| time::strptime(v, "%A, %d-%b-%y %H:%M:%S %Z"))
-                    .or_else(|_| time::strptime(v, "%a, %d-%b-%Y %H:%M:%S %Z"))
-                    .or_else(|_| time::strptime(v, "%a %b %d %H:%M:%S %Y"));
+                let tm = parse_gmt_date(v, "%a, %d %b %Y %H:%M:%S GMT")
```
SergioBenitez: This is wrong. It will fail to parse time zones that aren't GMT as allowed by RFC 822.

SergioBenitez:
@djc @jhpratt Thank you for you contributions here.

src/builder.rs
jhpratt: With regard to the alloc feature, I agree it's absolutely an incorrect usage. I originally had it as a "std" feature, but serde has an alloc feature, which was necessary to enable for 1.34 support. So it's actually an upstream issue from my end, and one that was unavoidable with the mandated 1.34 support.

If 1.36 were the MSRV, I could use the alloc crate unconditionally.

src/lib.rs
jhpratt: A year varies in length, unlike the rest of the lengths (we assume leap seconds aren't a thing).

Once tzdb support is implemented, there will be a `Period` type that will have a `year` method. Since `Duration`s boil down to seconds, it wouldn't be possible to just say "add one year" until a `Period` type exists (necessitating the tzdb).

src/parse.rs
jhpratt: The status quo is that `%Z` will completely ignore the provided time zone if it's neither "GMT" nor "UTC", however. That's a very subtle thing that can cause errors and is not documented directly in v0.1.

src/lib.rs
jhpratt: Is this not what is required for the `Expires` field? A quick search seems to indicate just that, ]at least from MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date).

jhpratt:
I took a quick look at the required format and believe that you're not remembering the required format correctly (not that I blame you). The `alloc` feature is a complete oddity, and was only changed recently due to the same feature in serde. I actually will raise an issue there to see if it's possible to fix the behavior, allowing me to do the same.

Thanks for the review! I'm definitely going to continue work on time. Hopefully it'll be in decent shape soon :slightly_smiling_face:

src/builder.rs
jhpratt: Also of note: `OffsetDateTime` implements all comparisons (equality, sorting, hashing, etc) in terms of its unix timestamp, so the only thing the offset is actually used for is formatting.

SergioBenitez: I was incorrect in stating that the MSRV is 1.34: it is, in fact, 1.36, as ring also makes use of `alloc`.

src/lib.rs
jebrosen: As far as I can find (and briefly discussed on IRC) this is the relevant point in RFC <https://tools.ietf.org/html/rfc2616#section-3.3.1>:

> All HTTP date/time stamps MUST be represented in Greenwich Mean Time
>   (GMT), without exception. For the purposes of HTTP, GMT is exactly
>   equal to UTC (Coordinated Universal Time).

src/lib.rs
SergioBenitez: That's fair and sensible. Thanks for explaining that.

src/builder.rs
jhpratt: In which case I could revert to a `std` feature as it was before. I'll hold off on doing that unless this is the final thing, though, since I hate going back-and-forth on feature names.

src/lib.rs
SergioBenitez: ...which is referred to by https://tools.ietf.org/html/rfc6265#section-4.1. Indeed, it seems that it is entirely correct, and even required, to be converted to GMT!

src/parse.rs
djc: The spec requires GMT anyway, so it seems okay to require GMT here.

SergioBenitez:
Merged in 490e5324aae011a74e877107ce348f7a423392ec and d35d8e4355b47ce6fa5289ef9af052346e3efaef. Thank you, again, and apologies for the misunderstanding on my behalf!

PR #137, opened 2020-01-29
Allow filling CookieJars from a Cookie header string

groomble:
I needed this ability so that I could use this library with Hyper. Contemporary Hyper has no cookie parsing logic, so the ability to parse a `&str` into a CookieJar is very helpful for making stateful manipulations of cookies. Since this is only useful for a `&str` with a lifetime that isn't `'static`, I needed CookieJars to be able to hold Cookies with lifetimes that aren't `'static`, hence the sweeping addition of lifetimes to those structs. (thank you very much for the no-allocation parsing!)

The actual method is at `src/jar.rs:126`.

I did not bump the version, as I'm not quite sure how this interacts with semver requirements; I didn't have to modify anything in the test suite, but I can see how the addition of lifetime parameters might be a breaking change (e.g. if a SignedJar was a struct field).

Feel free to mention any tasks I can do to help merge this! I'm still fairly new to Rust so let me know if there's anything to improve.

### DIFF
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -1,5 +1,6 @@
 use std::collections::HashSet;
 use std::mem::replace;
+use parse::ParseError;
 
 use time::{Duration, OffsetDateTime};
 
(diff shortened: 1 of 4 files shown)


src/jar.rs
```diff
@@ -96,10 +97,40 @@ impl CookieJar {
...
+    /// // name2, value2
+    /// // name3, value3
+    /// for cookie in jar.iter() {
+    ///     let (name, value) = cookie.name_value();
+    ///     println!("{}, {}", name, value);
+    /// }
+    /// ```
+    pub fn parse(header: &str) -> Result<CookieJar, ParseError> {
+        let mut ret = CookieJar::default();
+        for cookiestr in header.split("; ") {
```
SergioBenitez: This should split on `';'`, the `;` character, not the `"; "` string. Note, especially, the space. It should also handle arbitrary spacing around cookie values. Here's what we do in Rocket, for instance:

```rust
for cookie_str in raw_str.split(';').map(|s| s.trim()) {
    ...
}
```

SergioBenitez:
I'm all for the lifetime changes, but the inclusion of the `parse()` method has issues similar to ones in the previous PR to attempt to add this: https://github.com/SergioBenitez/cookie-rs/pull/107#issuecomment-575771128.

rsnakard-r7:
Commenting because I'm interested in this functionality 

> The functionality is far too subjective, in my opinion, for inclusion in the library. For instance, this implementation assumes at least the following:
> 
> 1. You want to add the cookies with add and not add_original.
> 1. You want to keep the latter of a duplicate if one exists.
> 1. You don't want to ignore parsing errors as you go.
> 1. You want the cookies in a regular CookieJar.

1. This solution creates a new CookieJar thus `add_original` should be used.
1. See 1. 
1. There are so many ways to handle and not handle errors this is going to be a problem with any collection. Would you be open to a `parse_list` function on Cookie that returns a `Iterator<Result<Cookie>>`?
1. I don't understand why this is a problem. What's wrong with `CookieJar::parse(s).private()` or `CookieJar::parse(s).signed()`?

SergioBenitez:
Closing due to inactivity.

PR #141, opened 2020-02-04
Update parse to match RFC 6265 Grammar

rsnakard-r7:
RFC 6265 section 4.1.1 defines the cookie-value as an octet that can be
surrounded by double quotes. Thus surrounding double quotes are not part
of the cookie-value and should be stripped during parsing.

ISSUE: 140

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -112,7 +112,7 @@ fn parse_inner<'c>(s: &str, decode: bool) -> Result<Cookie<'c>, ParseError> {
     // Determine the name = val.
     let key_value = attributes.next().expect("first str::split().next() returns Some");
     let (name, value) = match key_value.find('=') {
-        Some(i) => (key_value[..i].trim(), key_value[(i + 1)..].trim()),
+        Some(i) => (key_value[..i].trim(), key_value[(i + 1)..].trim_matches(|c: char| c.is_whitespace() || c == '\"')),
         None => return Err(ParseError::MissingPair)
     };
 
@@ -321,7 +321,9 @@ mod tests {
         let mut expected = Cookie::build("foo", "bar").finish();
         assert_eq_parse!("foo=bar", expected);
         assert_eq_parse!("foo = bar", expected);
+        assert_eq_parse!("foo=\"bar\"", expected);
         assert_eq_parse!(" foo=bar ", expected);
+        assert_eq_parse!(" foo=\"bar\" ", expected);
         assert_eq_parse!(" foo=bar ;Domain=", expected);
         assert_eq_parse!(" foo=bar ;Domain= ", expected);
         assert_eq_parse!(" foo=bar ;Ignored", expected);
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -20,6 +20,8 @@ pub enum ParseError {
     MissingPair,
     /// The cookie's name was empty.
     EmptyName,
+    /// The cookie contained invalid characters
+    InvalidChar(char),
     /// Decoding the cookie's name or value resulted in invalid UTF-8.
     Utf8Error(Utf8Error),
     /// It is discouraged to exhaustively match on this enum as its variants may
@@ -34,6 +36,7 @@ impl ParseError {
         match *self {
             ParseError::MissingPair => "the cookie is missing a name/value pair",
             ParseError::EmptyName => "the cookie's name is empty",
+            ParseError::InvalidChar(_) => "the cookie contained invalid characters",
             ParseError::Utf8Error(_) => {
                 "decoding the cookie's name or value resulted in invalid UTF-8"
             }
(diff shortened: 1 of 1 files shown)


rsnakard-r7:
https://github.com/SergioBenitez/cookie-rs/issues/140

SergioBenitez:
While I agree that the cookie value can include surrounding double-quotes, it's not obvious to me that the double quotes are _not_ part of the cookie value. Grammatically speaking, the double quotes _are_ part of the cookie value, so this is a question of semantics. Unfortunately, RFC 6265 explicitly states: "The semantics of the cookie-value are not defined by this document."

I'm not necessarily opposed to this change (though currently, the PR removes any number of quotes from either side when, according to the PR text's reference to the RFC, it should remove one quote from each side if one exists on both sides), but the motivation provided isn't sufficient. What if the surrounding double quotes are meaningful for consumers of the cookie value? This PR would make determining whether a value is surrounded by double quotes imperceptible.

To merge this, we'll need support in the form of an RFC defining these semantics for higher-level consumers or other libraries with significant users functioning this way.

rsnakard-r7:
Removing the wrapped quotations is the behavior of Java Spring, Express.js, Django, and Rails

SergioBenitez:
@rsnakard-r7 Can you point me to docs/code in each of the frameworks you listed?

rsnakard-r7:
**Cookie:**
Strips quotes:          https://github.com/jshttp/cookie/blob/master/index.js#L71
Used by express.js: https://github.com/expressjs/cookie-parser/blob/master/index.js#L15

**Django**
Strips quotes: https://github.com/django/django/blob/master/django/http/cookie.py#L25

**Apache Tomcat (used by Spring)**
Apache Tomcat has a cookie parser written for 6265 compliance
Strips quotes: https://github.com/apache/tomcat/blob/master/java/org/apache/tomcat/util/http/parser/Cookie.java#L491

**MDN** - Emphasis Mine
> A \<cookie-value\> can optionally be wrapped in double quotes and include any US-ASCII characters **\*\*excluding\*\*** control characters, Whitespace, **\*\*double quotes\*\***, comma, semicolon, and backslash.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Directives

src/parse.rs
```diff
@@ -102,6 +102,15 @@ fn name_val_decoded(
     unreachable!("This function should never be called with 'percent-encode' disabled!")
 }
 
+fn trim_quotes<'a>(s: &'a str) -> &'a str {
```
SergioBenitez: If the length of the string is `1`, this function will panic.

SergioBenitez:
Thank you! Convinced beyond a doubt.

Unfortunately the implementation provided is incorrect; it will panic on strings of `"`.

rsnakard-r7:
My first thought is to wrap trim_quotes in a `if len > 1`. Do you have a cleaner solution?

SergioBenitez:
I've updated my inline comment with a suggested implementation.

SergioBenitez:
Wonderful. Thank you! Merged in fcdb8b5d2e2875bb643c11b37b389e28f4717751.

mrobinson:
I think this change may cause Servo's cookie unit tests to fail and blocking the upgrade to the newest version of `cookie-rs`.  For background, our tests are from https://github.com/abarth/http-state, which has the same author as RFC 6265. Here's an example of a test in question that expects surrounded quotes to be preserved as a part of the cookie value: https://github.com/abarth/http-state/blob/master/tests/data/parser/chromium0007-test. I'm fairly sure that in addition to Chromium, Firefox is preserving these quotes as well.

Reading the RFC, it's unclear to me why the outer quotes should be stripped as they are explicitly included as part of the value:

```
cookie-value      = *cookie-octet / ( DQUOTE *cookie-octet DQUOTE )
cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                       ; US-ASCII characters excluding CTLs,
                       ; whitespace DQUOTE, comma, semicolon,
                       ; and backslash
```

Could it be that the tools that strip quotes are misinterpreting this section of the RFC which suggests that the outer quotes should be preserved, but inner ones should be stripped?

soloturn:
@SergioBenitez ?

SergioBenitez:
I think the only solution here, given that there appears to be a wide disagreement on the issue in the wild (though I agree that the RFC is quite clear here), is to provide implementations that do _and_ don't strip the quotes. I'll work on that now.

SergioBenitez:
I've pushed https://github.com/SergioBenitez/cookie-rs/commit/89edddf7152c263c03bf19338a571a01f971cf34 which hopefully addresses this for good. I'm working on a few other quality-of-life breaking changes, then I'll push a release with everything included.

mrobinson:
@SergioBenitez Thanks very much!

PR #144, opened 2020-04-02
Use HTTPS

atouchet:


SergioBenitez:
These all redirect to their HTTPS counterparts already.

atouchet:
Is there a reason why you wouldn't want to link to the HTTPS URL directly?

SergioBenitez:
No reason other than to prevent a superfluous change. The proposed change and the current revision are functionally identical.

PR #146, opened 2020-04-12
Add domain to hash function of DeltaCookie

saskenuba:
Since cookies were hashed by their names, cookies with the same name but different domains were getting overwritten. 
If the cookie has a domain, we should use it to create a hash, or is there some reason not to?

Thanks!

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -61,6 +61,9 @@ impl Eq for DeltaCookie {}
 impl Hash for DeltaCookie {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.name().hash(state);
+        if let Some(domain) = self.domain() {
+            domain.hash(state)
+        }
     }
 }
 


SergioBenitez:
I'm not necessarily opposed to a change in this spirit, but such a change would need quite a bit more consideration. At the moment, all of the APIs and logic are constructed around the idea that a cookie's `name` uniquely identifies a cookie: it's how you get cookies from a jar, remove them, and so on. Effectively, it's how equality of two cookies is defined.

To make the change proposed in this PR is to change the definition of equality for cookies in the crate in its entirely. At minimum, this means that we'd need to make the following changes:

- The `get`/`remove` (etc.) `CookieJar` APIs would need to be reworked to allow specifying domains and paths. The following questions would need to be answered:
  - Do we preserve the ability to query cookies by name only? If so, what does `get` return when there is more than one cookie with the same name? What does `remove` do in the same situation?
  - What API do we expose to make the simple name-only case simple?
- How should replacement work, say, with `add_original`?
- We'd want to distinguish by `path` as well.
- `PartialEq` would need to be updated.
- All relevant documentation would need to be updated.

SergioBenitez:
Closing due to inactivity.

PR #149, opened 2020-06-08
Update 'aes-gcm', 'hmac', 'hkdf', 'sha2'

tarcieri:
This release uses the new `aead` v0.3 crate which makes the following changes to the API:

- `NewAead` now borrows the key, preventing needless copies
- `Aead` and `AeadInPlace` traits have been split. It appears you're using both, so this pulls them both in.

Note that these crates now depend on `generic-array` v0.14 which has an MSRV of 1.41+.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ time = { version = "0.2.11", default-features = false, features = ["std"] }
 percent-encoding = { version = "2.0", optional = true }
 
 # dependencies for secure (private/signed) functionality
-aes-gcm = { version = "0.5.0", optional = true }
+aes-gcm = { version = "0.6.0", optional = true }
 hmac = { version = "0.7.1", optional = true }
 sha2 = { version = "0.8.2", optional = true }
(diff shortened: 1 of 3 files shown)


tarcieri:
Note: we're planning a similar update to the `digest` crate soon (i.e. later this week) as well as the `sha2` crate which will also bump `generic-array` up to v0.14 there. Just a heads up that if you merge this without that change, you'll be depending on two different versions of `generic-array`.

SergioBenitez:
This is awesome, @tarcieri! Thank you! I'm hoping https://github.com/fizyk20/generic-array/pull/102 is merged before moving ahead.

tarcieri:
Ooh, neat PR to `generic-array`. Good to know!

tarcieri:
Looks like `generic-array` v0.14.2 is out with your PR included?

SergioBenitez:
Awesome. Merged in f7f6e741521c866bf9b6c290b4389425b5c952e7! Thank you!

PR #152, opened 2020-06-15
add plain cookie wrappers

robjtede:
Further to comment discussions on https://github.com/actix/actix-web/pull/1558.

Will fill in remaining docs if this approach is acceptable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,6 +87,7 @@ mod parse;
 mod jar;
 mod delta;
 mod draft;
+mod plain;
 
 #[cfg(any(feature = "private", feature = "signed"))] #[macro_use] mod secure;
 #[cfg(any(feature = "private", feature = "signed"))] pub use secure::*;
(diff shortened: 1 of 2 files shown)


rsnakard-r7:
Why not use `Cookie::name_value()`?

robjtede:
That could help with the plain case but not the plain_encoded case.

robjtede:
Interested to hear your view on what this API should look like @SergioBenitez given you've seen the use-case we're trying to achieve.

SergioBenitez:
@robjtede Apologies for the delay. I've revamped the way configurable displaying works in https://github.com/SergioBenitez/cookie-rs/commit/7efcdd83b8ded99ea86a9d00ed72cd211323949b. You can do:

```rust
cookie.encoded(); // encode name/value, display everything, as before
cookie.stripped(); // display only name/value, unencoded
cookie.stripped().encoded(); // display encoded name/value
cookie.encoded().stripped(); // display encoded name/value
```

This will ship in the next major release.

PR #156, opened 2020-06-22
Create parse_string method that returns an iterator over cookies

rsnakard-r7:
The motivation for this API is that I don't want to iterate over a cookie string myself if I can entrust it to a library. This function returns an iterator over cookie Results. This allows the implementer to add each cookie to their cookie jar as either deltas or original cookies. It also allows them to handle errors however they choose.

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -3,46 +3,41 @@ use crate::parse::parse_cookie;
 use crate::parse::ParseError;
 use std::borrow::Cow;
 
+/// An iterator over a list of `cookie`s. Follows the HTTP request cookie spec. Accepts a semicolon separated list of key-value pairs.
+/// 
+/// This struct is created by the `parse_string` method. See its documentation for more
 pub struct CookieIter<'c> {
-    pub input: Option<Cow<'c, str>>,
-    pub decode: bool,
+    /// The remaining input that has not yet been iterated over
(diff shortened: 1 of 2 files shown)


rsnakard-r7:
@SergioBenitez is this something you would think about adding to Cookie?

SergioBenitez:
> @SergioBenitez is this something you would think about adding to Cookie?

Yes. However, this seems much more complicated than it need be. Why not something like:

```rust
raw_str.split(';')
    .map(|s| s.trim())
    .filter(|s| !s.is_empty())
    .map(Cookie::parse) // or `parse_encoded`
```

What's more, if we implement `FromIterator` for `CookieJar` (likely calling `add_original`), we recover the original intent in #137 with a much more flexible solution:

```rust
let jar: Result<CookieJar, _> = CookieJar::split(foo).collect();
```

In terms of method names, I think maybe `split()/split_encoded()` or `parse()/parse_encoded()` would be preferred; or perhaps three: `split()/parse()/parse_encoded()`, where `split()` does everything except call `Cookie::parse`.

rsnakard-r7:
I don't see how we can implement FromIterator without running into the problems in https://github.com/SergioBenitez/cookie-rs/pull/107#issuecomment-575771128

PR #160, opened 2020-07-21
src: run cargo fmt

Fishrock123:
I accidentally ran this in a more important PR... This should be added to ci probably...


SergioBenitez:
I don't believe we should run `cargo fmt` on the codebase. See https://github.com/SergioBenitez/Rocket/pull/1211#issuecomment-656052249 for more.

PR #161, opened 2020-07-21
SignedJar: add key-rotatable version

Fishrock123:
Implements https://github.com/SergioBenitez/cookie-rs/issues/158

Maybe this should be the default (`::new()`) in the future.

### DIFF
--- a/src/secure/signed.rs
+++ b/src/secure/signed.rs
@@ -19,7 +19,8 @@ pub(crate) const KEY_LEN: usize = 32;
 #[cfg_attr(all(doc, not(doctest)), doc(cfg(feature = "signed")))]
 pub struct SignedJar<'a> {
     parent: &'a mut CookieJar,
-    key: [u8; KEY_LEN],
+    rotated_keys: Vec<[u8; KEY_LEN]>, // Older rotated keys.
+    key: [u8; KEY_LEN],               // The primary (newest) key.
 }
 
 impl<'a> SignedJar<'a> {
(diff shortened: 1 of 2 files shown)


src/secure/signed.rs
```diff
@@ -57,9 +73,24 @@ impl<'a> SignedJar<'a> {
...
+        if mac.verify(&digest).is_ok() {
+            return Ok(value.to_string());
+        }
+
+        println!("First key not match");
+
+        for key in &self.rotated_keys {
+            let mut mac = Hmac::<Sha256>::new_varkey(key).expect("good key");
+            mac.update(value.as_bytes());
+            // if mac.verify(&digest).is_ok() {
```
Fishrock123, line 85: Oops, I forgot ti switch this back, fixing...

SergioBenitez:
I quite like the idea - it's something I'd been desiring for some time.

My take is that rotation should be a feature of the `Key` structure, not of individual child jars. This likely means that `Key` becomes an enum, or that we create two `~Key` structures that implement a unifying trait. It would be nice to have a `shift()` method on `RotatingKey` (or whatever you want to call it) that pushes in a new key; it would also be nice if you could set a maximum number of keys that `RotatingKey` kept around.

The `trait` approach would be quite nice as it would allow the jars to be totally unaware of where the key material is coming from or whether rotation is occurring. The trait/impls might look like:

```rust
trait Key {
    fn from(key: &[u8]) -> Self;
    fn derive_from(material: &[u8]) -> Self;
    fn try_generate() -> Option<Key>;
    fn generate() -> Option<Key> {
        Self::try_generate().expect("failed to generate key")
    }

    fn signing(&self) -> &[u8];
    fn encryption(&self) -> &[u8];
    fn master(&self) -> &[u8];
}

struct StaticKey { .. }
impl Key for StaticKey { .. }

struct RotatingKey { .. }
impl Key for RotatingKey { .. }
```

I _think_ it would then make sense for the child jars to keep around a dynamic trait object:

```rust
struct PrivateJar<'k, 'j> {
    key: &'k dyn Key,
    jar: &'j CookieJar<'static>
}
```

SergioBenitez:
Closing due to inactivity, but please note that I'm generally in favor of this.

PR #170, opened 2020-12-26
Update 'rand' to '0.8'

paolobarbolini:


SergioBenitez:
Sincerest apologies. I totally missed this PR when doing some work on `cookie` today and accidentally made the same change. Thank you, however!

PR #174, opened 2021-05-03
Update RustCrypto dependencies

tarcieri:
Updates the following RustCrypto dependencies to the latest versions:

- `aes-gcm` v0.9.0
- `hmac` v0.11.0
- `hkdf` v0.11.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,12 +25,12 @@ time = { version = "0.2.11", default-features = false, features = ["std"] }
 percent-encoding = { version = "2.0", optional = true }
 
 # dependencies for secure (private/signed) functionality
-aes-gcm = { version = "0.8.0", optional = true }
-hmac = { version = "0.10.0", optional = true }
+aes-gcm = { version = "0.9.0", optional = true }
+hmac = { version = "0.11.0", optional = true }
 sha2 = { version = "0.9.0", optional = true }
 base64 = { version = "0.13", optional = true }
 rand = { version = "0.8", optional = true }
-hkdf = { version = "0.10.0", optional = true }
+hkdf = { version = "0.11.0", optional = true }
 subtle = { version = "2.3", optional = true }
 
 [build-dependencies]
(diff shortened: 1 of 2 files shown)


tarcieri:
Heads up: the [RUSTEC-2021-0059](https://rustsec.org/advisories/RUSTSEC-2021-0059.html) and [RUSTSEC-2021-0060](https://rustsec.org/advisories/RUSTSEC-2021-0060.html) unmaintained crate advisories will likely apply to current users until these are updated, in case you start getting reports.

SergioBenitez:
Thank you for keeping the libs updates. I really appreciate your effort here.

Merged in 14b053a609eb494a09020f31f327a71c0a525dec.

SergioBenitez:
@tarcieri It looks like `aes` has a new MSRV. Previously, it compiled just fine on `1.41`. Now, I'm getting the following:

```rust
error[E0658]: unions with non-`Copy` fields are unstable
   --> aes-0.7.0/src/autodetect.rs:30:13
    |
30  | /             pub(super) union Inner {
31  | |                 pub(super) ni: ManuallyDrop<crate::ni::$name>,
32  | |                 pub(super) soft: ManuallyDrop<crate::soft::$name>,
33  | |             }
    | |_____________^
...
125 |   define_aes_impl!(Aes128, aes128, U16, "AES-128 block cipher instance");
    |   ----------------------------------------------------------------------- in this macro invocation
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/55149
```

Any chance MSRV could be pushed back to 1.41?

tarcieri:
@SergioBenitez with the `force-soft` feature, the MSRV of the `aes` crate is still 1.41+.

I now realize there's no corresponding `force-soft` feature for the `aes-gcm` crate itself though. I could potentially add that if you'd like to keep MSRV at 1.41.

SergioBenitez:
> @SergioBenitez with the `force-soft` feature, the MSRV of the `aes` crate is still 1.41+.
> 
> 
> 
> I now realize there's no corresponding `force-soft` feature for the `aes-gcm` crate itself though. I could potentially add that if you'd like to keep MSRV at 1.41.

While it's debatable whether MSRV should be part of semver or not, I'd like to be conservative here and not break MSRV in a minor point release, so this would be ideal.

PR #176, opened 2021-05-05
Bump `aes-gcm` to v0.9.1 and enable `force-soft`; MSRV 1.41+

tarcieri:
I released a new version of `aes-gcm` with a `force-soft` feature:

https://github.com/RustCrypto/AEADs/pull/306

This disables autodetection of AES-NI, falling back on a portable constant-time software implementation.

However, in this configuration `aes-gcm` still supports an MSRV of 1.41+, as opposed to 1.49+ required by the autodetection feature.

A comment in Cargo.toml notes `force-soft` should be removed when MSRV is bumped to at least 1.49.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ time = { version = "0.2.11", default-features = false, features = ["std"] }
 percent-encoding = { version = "2.0", optional = true }
 
 # dependencies for secure (private/signed) functionality
-aes-gcm = { version = "0.9.0", optional = true }
+aes-gcm = { version = "0.9.1", optional = true, features = ["force-soft"] } # remove `force-soft` when MSRV 1.49+
 hmac = { version = "0.11.0", optional = true }
 sha2 = { version = "0.9.0", optional = true }
 base64 = { version = "0.13", optional = true }


SergioBenitez:
> This disables autodetection of AES-NI, falling back on a portable constant-time software implementation.

Did `aes` take advantage of AES-NI before on stable? If so, is there any way to regain this support?

tarcieri:
@SergioBenitez it previously would use AES-NI only if the requisite target flags were explicitly enabled with:

```
RUSTFLAGS=-Ctarget-feature=+aes,+ssse3
```

The v0.9 release adds runtime autodetection on `i686`/`x86-64` targets, checking CPUID for AES-NI (and CLMUL) availability and automatically selecting the AES-NI/CLMUL backends with no explicit configuration by the user.

Unfortunately, the strategy we used to implement this depends on some features introduced in Rust 1.49 (namely support for non-`Copy` unions using `ManuallyDrop`).

Unless the `RUSTFLAGS` were explicitly configured to enable these target features, however, there will be no user-facing change to using the `force-soft` feature.

SergioBenitez:
Got it. I suspect `-C target-cpu=native` is a much more common setting that would have enabled AES-NI previously. Does static detection supersede dynamic detection even now? That is, if I compile with `-C target-cpu=native`, can I avoid the cost of dynamically checking for AES-NI on `0.9` _and_ still use AES-NI? This seems like a good compromise to me.

tarcieri:
Yes, if the requisite target features are enabled, the runtime check compiles down to a noop and only the AES-NI implementation will be linked into the resulting executable.

Unfortunately, though that's a nice optimization which we get for free today, the code is still written in a way that requires Rust 1.49+.

SergioBenitez:
Am I understanding correctly that this is a monotonic regression? Even if I compile with `-C target-cpu=native`, setting `force-soft` will result in AES-NI _not_ being used, even though its presence can be statically determined, is that correct? If so, this feels rather unfortunate.

tarcieri:
Yes, using `force-soft` will disable AES-NI support completely. I would only suggest it as a temporary stopgap until you can bump MSRV 1.49+.

If you'd like to support AES-NI on MSRV 1.41+, I'd suggest sticking with `aes-gcm` v0.8 until you can bump MSRV.

SergioBenitez:
> Yes, using `force-soft` will disable AES-NI support completely.

This seems like an ill-fitted use of features. Features should only be used to _add_ capabilities, but `force-soft` _removes_ the ability to make use of AES-NI. The consequence is that a consumer of `aes` has little control over whether AES-NI is actually used: all it takes is a single dependency, direct or indirect, enabling `force-soft` to disable AES-NI completely, without recourse.

I would suggest what I would consider to be and have been the "correct" approach: _always_ enable AES-NI if it is known to statically exist, and otherwise, add a feature that adds the ability to dynamically detect its existence, which of course becomes a no-op if the first condition is met. Ideally, the MSRV is only bumped if the feature is enabled, resolve two issues in one.

SergioBenitez:
I've taken a look at the issue you've linked, and I've read and reread your post, but I still don't understand where the technical difficulty lies in enabling AES-NI if either it is statically known to be available _or_ it can be detected to be available at runtime. Every other option feels remarkably wrong to me. For AES and SHA256, especially on x86, I _always_ expect hardware supported operations if they are available. If the library supports it, it should not be possible for some transitive dependency to disable such support, and ideally, I shouldn't need to upgrade compilers to obtain this support, either. I consider either a huge misstep. The fact that a version _upgrade_  results in losing important functionality makes the situation even less ideal.

> Unless we lean on the autodetection codepaths to do this, we now have two cases to deal with:
> 
> 1. `target-feature=+aes` is enabled, and `autodetect` is not
> 2. `target-feature=+aes` is enabled, and `autodetect` is also enabled
> 
> In our current scheme, we don't deal with the first case. We used to, and our code used to be riddled with a ton of `#[cfg(target_feature = "...")]` gating. If you look at the commits/PRs I linked with an `autodetect` feature they were using a lot of complex `cfg-if` driven gating for this purpose. Now all of that has been abstracted away into the `cpufeatures` crate.

Somewhere in the codebase, you already have 2 versions of the code. Let's call them `aes::ni` and `aes::soft`.

Effectively, we want this:

```rust
fn aes() {
    #[cfg(have_aesni)]
    return aes::ni();

    #[cfg(feature(autodetect)]
    if detect(aes-ni) { return aes::ni() }

    aes::soft()
}
```

This structure makes it clear that `autodetect` is _additive_; it _adds_ the ability to dynamically detect AES-NI and removes nothing otherwise; there is no `not(feature(autodetect))`. This is the only fully correct use of features. If the code that requires `1.49` is restricted to `detect`, then MSRV can remain at 1.41 as long as `autodetect` is not enabled.

> There is perhaps a case to be made for using something like:
> 
> ```toml
> [features]
> default = ["autodetect"]
> ```
> 
> ...but any Rust Embedded user can tell you about their fraught endeavors trying to chase down every `default = ["std"]` feature activation directive in all of their transitive dependencies.

Aren't you enabling auto-detection by default now? I don't see what the problem is either way. It's a simple line or two of documentation to say, "hey, this is how you use this crate in a `no-std` environment, and here's what you'll miss out on."

tarcieri:
> Somewhere in the codebase, you already have 2 versions of the code. Let's call them aes::ni and aes::soft.
>
> Effectively, we want this:

It's a lot more complicated than that...

https://github.com/RustCrypto/block-ciphers/blob/master/aes/src/autodetect.rs

Anyway, I'd suggest opening an upstream issue on the `block-ciphers` repo about this so yet another round of discussion about it is captured there.

SergioBenitez:
This is now in 0.16.

PR #196, opened 2022-01-05
Use base64ct instead of base64

### LINKED ISSUE #195: Replace base64 with base64ct
The [`base64ct`](https://github.com/RustCrypto/formats/tree/master/base64ct) crate is specifically designed to provide constant-time decoding and encoding of Base64 strings for cryptographic and security-sensitive applications. Since in this crate `base64` is used for handling sensitive data, it's probably worth to replace it with `base64ct`.

Unfortunately, using `base64ct` would mean MSRV bump to at least 1.51 (for pre-1.3 versions) or to 1.56 for the latest released version.

cc @tarcieri

newpavlov:
Closes #195.

Also applies 2018 idiom lints.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,8 +2,7 @@
 name = "cookie"
 version = "0.16.0"
 authors = ["Sergio Benitez <sb@sergio.bz>", "Alex Crichton <alex@alexcrichton.com>"]
-edition = "2021"
-rust-version = "1.56"
+edition = "2018"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/SergioBenitez/cookie-rs"
 documentation = "https://docs.rs/cookie"
@@ -29,7 +28,8 @@ percent-encoding = { version = "2.0", optional = true }
 aes-gcm = { version = "0.9.0", optional = true }
 hmac = { version = "0.12.0", optional = true }
 sha2 = { version = "0.10.0", optional = true }
-base64ct = { version = "1", features = ["alloc"], optional = true }
+# note: update after MSRV will be bumped to 1.56 or higher
+base64ct = { version = ">=1, <1.1.0", features = ["alloc"], optional = true }
 rand = { version = "0.8", optional = true }
 hkdf = { version = "0.12.0", optional = true }
 subtle = { version = "2.3", optional = true }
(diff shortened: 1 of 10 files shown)


SergioBenitez:
My initial inclinations here are:

  1. Why?
  2. Why not?

I say this half in jest. :)

Back to reality: I cannot discern a situation in which a timing of `base64` reveals security sensitive information. The `base64` value is fully known to any attacker, so the timing of a base64 decode of said value would also be known. The decoding operation does not make use of any sensitive information. This would imply that a proof of interference between sensitive information and the base64 values exists, and so variable-time decoding of base64 in this context is as secure as constant-time decoding of base64 in this context. Do you see it another way?

Still, if this is no answer to "why not?", there's no reason _not_ to do this. The standard response would be that there's a performance cost. Can you measure said cost and report back results? We're more interested in _passing_ verifications than we are failing, which will obviously be slower but also rare.

newpavlov:
In a security-related context it's better to use const-timing algorithms by default. It may be unnecessary in this particular case, but side-channel attacks can be quite surprising and non-trivial. For example, if a secret value is used repeatedly an attacker  acting under a different user may try to estimate it by measuring timings of responses to his requests (such attack is probably highly unpractical, but still theoretically possible).

Performance-wise, I got the following results on my PC:
```
test base64_decode   ... bench:      35,631 ns/iter (+/- 7,786) = 2104 MB/s
test base64_encode   ... bench:      33,999 ns/iter (+/- 1,025) = 2205 MB/s
test base64ct_decode ... bench:     123,016 ns/iter (+/- 6,242) = 609 MB/s
test base64ct_encode ... bench:      41,659 ns/iter (+/- 1,407) = 1800 MB/s
```
With `target-cpu=native`:
```
test base64_decode   ... bench:      34,659 ns/iter (+/- 288) = 2163 MB/s
test base64_encode   ... bench:      32,796 ns/iter (+/- 263) = 2286 MB/s
test base64ct_decode ... bench:      99,231 ns/iter (+/- 637) = 755 MB/s
test base64ct_encode ... bench:      28,777 ns/iter (+/- 353) = 2606 MB/s
```
It probably worth to look into decoding performance of `base64ct`, it may be possible to improve it (I did something similar for `base16ct` previously).

PR #201, opened 2022-07-14
Implemented TryFrom<&[u8]> for Key

e-rhodes:
The existing `Key::from` method, which creates a key from a `&[u8]`, panics when the key isn't long enough. This PR moves this logic into an impl of `TryFrom` that returns a `Result` instead of panicking. The existing `Key::from` now just unwraps the `TryFrom` impl, so its behavior is unchanged.

This allows better handling an improperly created key, as well as creating a `Key` without having `Key` in scope, e.g. when using another crate that requires a `Key` but doesn't reexport `cookie`.

### DIFF
--- a/src/secure/key.rs
+++ b/src/secure/key.rs
@@ -1,3 +1,5 @@
+use core::convert::{TryFrom, TryInto};
+
 const SIGNING_KEY_LEN: usize = 32;
 const ENCRYPTION_KEY_LEN: usize = 32;
 const COMBINED_KEY_LENGTH: usize = SIGNING_KEY_LEN + ENCRYPTION_KEY_LEN;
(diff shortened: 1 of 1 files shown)


SergioBenitez:
This is great, thanks for the contribution!

This is technically a [breaking change, albeit a minor one](https://github.com/rust-lang/rfcs/blob/721f2d74cc4daf76f3e49d58fbc6ded55d545e45/text/1105-api-evolution.md#minor-change-implementing-any-non-fundamental-trait), so perhaps we should wait for a major release. In any case, I'll merge this now; feel free to opine on the breakage considerations!

SergioBenitez:
> annotations mine. Since `TryFrom` is a std trait and it's parameterized by `&[u8]` (an std type), users couldn't create a conflicting implementation

This isn't quite the issue here. Consider the following:

```rust
struct Foo;

// implemented in `cookie`
impl std::convert::TryFrom<&[u8]> for Foo {
    type Error = ();

    fn try_from(_val: &[u8]) -> Result<Foo, ()> {
        println!("TryFrom");
        Ok(Foo)
    }
}

// in external crate
pub trait MyTrait<T>: Sized {
    type Error;

    fn try_from(value: T) -> Result<Self, Self::Error>;
}

// in external crate
impl MyTrait<&[u8]> for Foo {
    type Error = ();

    fn try_from(_val: &[u8]) -> Result<Foo, ()> {
        println!("MyTrait");
        Ok(Foo)
    }
}

// in external crate
fn main() {
    use std::convert::TryFrom;

    Foo::try_from(&[][..]); // ERROR: multiple applicable items in scope
}
```

e-rhodes:
Ahh, makes sense. I didn't consider that case. Thanks for taking the time!

Issue #109, opened 2019-01-09
time crate dependency


lnicola:
This might be a stupid question, but does the `time` crate offer anything in addition to `std::time`? Replacing it would be a breaking change, but is there more to it?

SergioBenitez:
In general, `time` on crates is a maintained/expanded version of `std::time`. It's got a bunch of helpful functions and methods that `std::time` doesn't, and it's generally agreed that one should use `time` over `std::time`.

lnicola:
> In general, time on crates is a maintained/expanded version of std::time.

That's not how I see it; the repository is at `rust-lang-deprecated/time` and the README says:

> This library is no longer actively maintained, but bugfixes will be added (details).

> it's generally agreed that one should use time over std::time.

Maybe, but I've never seen that advice.

> It's got a bunch of helpful functions

Got it, `strptime` and `strftime`. There's no equivalent functionality in `std::time`.

jhpratt:
@SergioBenitez `time`'s repository mentions that `chrono` should be used instead, as `time` is no longer maintained.

SergioBenitez:
@jhpratt Certainly, but `chrono` is a much bigger, much more opinionated crate than we need. That being said, I have no strong opinions on the matter. As long as it is user-friendly, I'm open to migration PRs.

jhpratt:
@onx2 I actually wound up taking over the time crate shortly after my previous comment. It's now a rough equal to chrono, so no reason to avoid one or the other.

onx2:
It seems like all the other packages I'm using prefer chrono or at least provide a feature for it. Would it be possible to add the option for chrono to this package? Or if that's something you don't want to do, would you suggest using both `time` and `chrono`?

jhpratt:
I'm not a maintainer, but I'm pretty sure that's something that wouldn't even be feasible. sqlx and async-graphql use the feature flags to enable third-party implementations whereas cookie relies on it in the API itself — basically you _couldn't_ have both enabled at the same time, and the features wouldn't be additive as required by cargo.

There's nothing wrong with using both crates, though I can imagine it may get a little confusing at times.

onx2:
Got it, thanks for the clarification! I'll use `time` for this crate and `chrono` for the others 😄

TheCGDF:
[...]

PR #213, opened 2023-09-15
`chrono` feature to allow library users to either use `time` or `chrono` dependency

bm-w:
The `chrono` feature brings in the [`chrono` crate][chrono_cio] as a dependency & uses it to reimplement all date & time functionality. In essence, it uses [`chrono::DateTime`][chrono_datetime]`<`[`chrono::FixedOffset`][chrono_datetime]`>` in `Expiration::DateTime` rather than [`time::OffsetDateTime`][time_offsetdatetime] (`FixedOffset` seems to best match `OffsetDateTime`’s `offset` field), and [`chrono::Duration`][chrono_duration] rather than [`time::Duration`][time_duration].

The `time` feature (which becomes a default feature) represents the library’s previous behavior, implementing date & time functionality using the [`time` crate][time_cio]. Gating it behind a feature lets users opt out of using that crate, and into using the [`chrono` crate][chrono_cio] instead.

In order to let the library continue working with `default-features = false`, all date & time functionality is gated behind either of these features. So too is some other functionality that relies on date & time functionality, such as “removal“ cookies. In order to avoid adding a lot of noise to doctests those too are gated behind the `time` feature.

This is a breaking change, given that it effectively removes all date & time functionality when neither feature is enabled, e.g. when `default-features = false`. Given that the `cookies` crate did not previously have default features, it seems plausible that few were actually disabling those. When `time` is enabled (which in many cases it will be as a default feature), none of the changes are breaking.

All regular tests pass, but I have not had the opportunity to run the fuzzers yet.

[chrono_cio]: https://crates.io/crates/chrono
[chrono_datetime]: https://docs.rs/chrono/0.4.30/chrono/struct.DateTime.html
[chrono_duration]: https://docs.rs/chrono/0.4.30/chrono/struct.Duration.html
[chrono_fixedoffset]: https://docs.rs/chrono/0.4.30/chrono/offset/struct.FixedOffset.html
[time_cio]: https://crates.io/crates/time
[time_duration]: https://docs.rs/time/0.3.28/time/struct.Duration.html
[time_offsetdatetime]: https://docs.rs/time/0.3.28/time/struct.OffsetDateTime.html

---

I went ahead an implemented this as I understood that it might be welcome from a [comment](https://github.com/SergioBenitez/cookie-rs/issues/109#issuecomment-524132207) by @SergioBenitez in #109. In another [comment](https://github.com/SergioBenitez/cookie-rs/issues/109#issuecomment-827071248), @jhpratt express concern that the features wouldn’t be additive, but IIUC strictly either feature on its own _is_ additive — they’re just _mutually exclusive_, for which [the Cargo Book carves out an exception](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).

On a final note, also discussed in #109, removing the use of `time::Duration` (or `chrono::Duration` for that matter) in favor of `std::time::Duration` would clean this PR up a bit: the `max_age` stuff wouldn’t need to be feature-gated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ default = ["time"]
 
 [dependencies]
 time = { version = "0.3", default-features = false, features = ["std", "parsing", "formatting", "macros"], optional = true }
-chrono = { version = "0.4.30", optional = true }
+chrono = { version = "=0.4.24", optional = true } # Version 0.4.25 bumps MSRV
 percent-encoding = { version = "2.0", optional = true }
 
 # dependencies for secure (private/signed) functionality
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,13 @@ secure = ["private", "signed", "key-expansion"]
 private = ["aes-gcm", "base64", "rand", "subtle"]
 signed = ["hmac", "sha2", "base64", "rand", "subtle"]
 key-expansion = ["sha2", "hkdf"]
+time = ["dep:time"]
+chrono = ["dep:chrono"]
+default = ["time"]
 
 [dependencies]
-time = { version = "0.3", default-features = false, features = ["std", "parsing", "formatting", "macros"] }
+time = { version = "0.3", default-features = false, features = ["std", "parsing", "formatting", "macros"], optional = true }
+chrono = { version = "0.4.30", optional = true }
 percent-encoding = { version = "2.0", optional = true }
 
 # dependencies for secure (private/signed) functionality
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -11,9 +11,21 @@ cargo test --verbose --features secure
 cargo test --verbose --features 'private,key-expansion'
 cargo test --verbose --features 'signed,key-expansion'
 cargo test --verbose --features 'secure,percent-encode'
+cargo test --verbose --no-default-features --features chrono
+cargo test --verbose --no-default-features --features 'chrono,private'
+cargo test --verbose --no-default-features --features 'chrono,signed'
+cargo test --verbose --no-default-features --features 'chrono,secure'
+cargo test --verbose --no-default-features --features 'chrono,private,key-expansion'
+cargo test --verbose --no-default-features --features 'chrono,signed,key-expansion'
+cargo test --verbose --no-default-features --features 'chrono,secure,key-expansion'
+
+if cargo test --verbose --features 'time,chrono' > /dev/null 2>&1; then
+    echo 'features \"time\" and \"chrono\" cannot be enabled at the same time' >&2
+    exit 1
+fi
 
 cargo test --verbose
-cargo test --verbose --no-default-features
-cargo test --verbose --all-features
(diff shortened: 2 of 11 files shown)


SergioBenitez:
There's no way this can work. The features need to be truly additive - there is no exception for crates that are used as transitive dependencies. Imagine a crate Foo that depends on crates A and B where A depends on Cookie with chrono and B depends on cookie with time. Foo will fail to compile, and there's no recourse.

The only way to add support for chrono and/or time is to create independent `cookie` time types that wrap either of the types and allow conversions to/from the types. When a feature is enabled, the respective conversions are implemented. Thus you can use both features at the same time. 

I'm not sure if a `T: Into<cookie::Duration>` bound would work, however. I'd be worried about the impls that appear based on features messing up inference. But we can try and test.

bm-w:
Understood and agreed.

I could try and see if I can make your suggestion work. However, if we’re already going the way of independent `cookie` time types, I think we should go all the way and remove all types from `time` crate from the public API. We’d still need either the `time` or `chrono` crate for parsing†, and we could gate conversions behind `time` & `chrono` features (modulo your concerns about inference).

_†As long as we have it handy we can and should use one of those crates for formatting as well, but we wouldn’t strictly need the crate, since `cookie` only ever formats string using [`parsing::FMT1`](https://github.com/SergioBenitez/cookie-rs/blob/v0.17/src/parse.rs#L19) which we could could conceivably do ourselves. Moot point, though._

SergioBenitez:
> I could try and see if I can make your suggestion work. However, if we’re already going the way of independent `cookie` time types, I think we should go all the way and remove all types from `time` crate from the public API. We’d still need either the `time` or `chrono` crate for parsing†, and we could gate conversions behind `time` & `chrono` features (modulo your concerns about inference).

I agree, though I still think we need to provide conversion impls and pick a crate, any crate (could even be `std::time`) to do the parsing for us. I don't think we want to get into the business of maintaining (and debugging...) time parsing or manipulation.

In short:

  * If we could back our internal types on `std::time`, that'd be great (though we might not be able to due to limitation of those types). And then by default we'd introduce no new dependencies. 
  * Provide two features: `time` and `chrono`, which enable conversions from the relevant types to our internal type.
  * Test whether we can safely have polymorphic functions like `fn max_age<T: Into<Duration>>(self, value: T)`, where `Duration` is the internal type and any of `time::Duration` and `crono::Duration` and `cookie::Duration` work as a `value`. Here, "safely" means that enabling a feature cannot prevent existing code from working. The danger is that enabling a feature enables a new impl which might mess with inference. We need to test/read.

jhpratt:
For what it's worth, I did offer to have a shared `Duration` type between `time` and `chrono` specifically for cases like this. The offer was rejected.

SergioBenitez:
It's not just `Duration`, unfortunately. We also need proper dates for expirations, and parsing of said dates.

SergioBenitez:
Closing in favor of moving forward with #217.

PR #217, opened 2023-09-23
`max_age::Duration` type to replace `time` crate’s.

### LINKED ISSUE #109: time crate dependency

bm-w:
This PR introduces a `max_age::Duration` type to use with [`Cookie::max_age`](https://docs.rs/cookie/0.17.0/cookie/struct.Cookie.html#method.max_age) (& related). It is part of an effort to resolve #109, and a building block towards making #213 truly additive.

I opted for a new type, rather than using [`std::time::Duration`](https://doc.rust-lang.org/1.72.1/std/time/struct.Duration.html), because the latter is just far larger (`u64` seconds instead of `u32`) and far more precise (nanoseconds) than what’s arguably needed for the purpose of cookies (see the new type’s docs for full motivation). @SergioBenitez, if you instead prefer `std::time::Duration` here I’m perfectly happy to make that change (I could add some conveniences like `from_mins` & `as_hours` in a `max_age::DurationExt` trait).

If we do go ahead with this `max_age::Duration` type I can add some more of the typical operator trait implementations (`Add`, `Sub`, etc.) and arithmetic methods (`checked_add`, `saturating_add`, etc.).

One downside of this change with respect to [`time::Duration`](https://docs.rs/time/0.3.28/time/struct.Duration.html#method.saturating_add) I’d like to note here is that the `set_max_age` methods that take an `impl Into<max_age::Duration>` no longer work with a plain `time::Duration` — instead of `.set_max_age(time_duration)` users would need to call e.g. `.set_max_age(Some(time_duration.into()))` (i.e. converting into `max_age::Duration` and wrapping in a `Some`).

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -14,8 +14,7 @@ use crate::{Cookie, SameSite, Expiration, max_age};
 ///
 /// ```rust
 /// # extern crate cookie;
-/// use cookie::Cookie;
-/// use cookie::max_age::Duration;
+/// use cookie::{Cookie, max_age::Duration};
 ///
 /// # fn main() {
 /// let cookie: Cookie = Cookie::build("name", "value")
@@ -88,8 +87,7 @@ impl<'c> CookieBuilder<'c> {
     ///
     /// ```rust
     /// # extern crate cookie;
-    /// use cookie::Cookie;
-    /// use cookie::max_age::Duration;
+    /// use cookie::{Cookie, max_age::Duration};
     ///
     /// # fn main() {
     /// let c = Cookie::build("foo", "bar")
@@ -205,8 +203,7 @@ impl<'c> CookieBuilder<'c> {
     ///
     /// ```rust
     /// # extern crate cookie;
-    /// use cookie::Cookie;
-    /// use cookie::max_age::Duration;
+    /// use cookie::{Cookie, max_age::Duration};
     ///
     /// # fn main() {
     /// let c = Cookie::build("foo", "bar")
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 
-use crate::{Cookie, SameSite, Expiration};
+use crate::{Cookie, SameSite, Expiration, max_age};
 
 /// Structure that follows the builder pattern for building `Cookie` structs.
 ///
@@ -15,7 +15,7 @@ use crate::{Cookie, SameSite, Expiration};
 /// ```rust
 /// # extern crate cookie;
 /// use cookie::Cookie;
-/// use cookie::time::Duration;
+/// use cookie::max_age::Duration;
 ///
 /// # fn main() {
 /// let cookie: Cookie = Cookie::build("name", "value")
@@ -23,7 +23,7 @@ use crate::{Cookie, SameSite, Expiration};
 ///     .path("/")
 ///     .secure(true)
 ///     .http_only(true)
-///     .max_age(Duration::days(1))
+///     .max_age(Duration::from_naive_days(1))
 ///     .finish();
 /// # }
 /// ```
@@ -89,18 +89,18 @@ impl<'c> CookieBuilder<'c> {
     /// ```rust
     /// # extern crate cookie;
     /// use cookie::Cookie;
-    /// use cookie::time::Duration;
+    /// use cookie::max_age::Duration;
     ///
     /// # fn main() {
     /// let c = Cookie::build("foo", "bar")
-    ///     .max_age(Duration::minutes(30))
+    ///     .max_age(Duration::from_mins(30))
     ///     .finish();
     ///
-    /// assert_eq!(c.max_age(), Some(Duration::seconds(30 * 60)));
+    /// assert_eq!(c.max_age(), Some(Duration::from_secs(30 * 60)));
     /// # }
     /// ```
     #[inline]
-    pub fn max_age(mut self, value: time::Duration) -> Self {
+    pub fn max_age(mut self, value: max_age::Duration) -> Self {
         self.cookie.set_max_age(value);
         self
     }
@@ -206,14 +206,14 @@ impl<'c> CookieBuilder<'c> {
     /// ```rust
     /// # extern crate cookie;
     /// use cookie::Cookie;
-    /// use cookie::time::Duration;
+    /// use cookie::max_age::Duration;
     ///
     /// # fn main() {
     /// let c = Cookie::build("foo", "bar")
     ///     .permanent()
     ///     .finish();
     ///
-    /// assert_eq!(c.max_age(), Some(Duration::days(365 * 20)));
+    /// assert_eq!(c.max_age(), Some(Duration::from_naive_days(365 * 20)));
     /// # assert!(c.expires().is_some());
     /// # }
     /// ```
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -246,7 +246,6 @@ impl CookieJar {
     /// ```rust
     /// # extern crate cookie;
     /// use cookie::{CookieJar, Cookie};
-    /// use cookie::time::Duration;
     ///
     /// # fn main() {
     /// let mut jar = CookieJar::new();
--- a/src/jar.rs
+++ b/src/jar.rs
@@ -187,7 +187,7 @@ impl CookieJar {
     /// ```rust
     /// # extern crate cookie;
     /// use cookie::{CookieJar, Cookie};
-    /// use cookie::time::Duration;
+    /// use cookie::max_age::Duration;
     ///
     /// # fn main() {
     /// let mut jar = CookieJar::new();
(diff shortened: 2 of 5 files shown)


SergioBenitez:
Hey! Thanks for doing this. I'm really looking forward to getting something like this merged.

> I opted for a new type, rather than using [`std::time::Duration`](https://doc.rust-lang.org/1.72.1/std/time/struct.Duration.html), because the latter is just far larger (`u64` seconds instead of `u32`) and far more precise (nanoseconds) than what’s arguably needed for the purpose of cookies (see the new type’s docs for full motivation). @SergioBenitez, if you instead prefer `std::time::Duration` here I’m perfectly happy to make that change (I could add some conveniences like `from_mins` & `as_hours` in a `max_age::DurationExt` trait).

While it's nice to optimize for size, I don't think it's a strong enough argument for us to reimplement functionality available elsewhere. While this type might be relatively simple to reimplement, implementing `DateTime` functionality will be anything but, and in general I consider implementing anything time-related outside the scope of this crate. And if we use an external crate to implement our internal `DateTime`, then we win nothing by implementing `Duration` ourselves.

My suggestion would be to pick one of `chrono` or `time` and build our types around those. IE:

```rust
pub mod time {
    pub struct DateTime(chrono::DateTime<chrono::Utc>);

    pub struct Duration(chrono::Duration);

    #[cfg(feature = "chrono")]
    impl<T> From<chrono::DateTime<T>> for DateTime<T> { .. }

    #[cfg(feature = "time")]
    impl From<time::OffsetDateTime> for DateTime<T> { .. }
}
```

Which one we pick I'm open to. Some ways to choose are:
  * Look at `cookie` dependents. Are they more often depending on `time` or `chrono`?
  * See which one is being used more in the wild (dependents, downloads/time, big crates, etc.).
  * See which one results in `cookie` compiling faster.
  * See which one makes our implementation more robust/simpler.

My guess is that the answer to most of these is `chrono`, except for the final one where you'd need to take a closer look. If it's clear one or the other makes our implementation significantly more simple or less error-prone (i.e, consider panics - we definitely want to _not_ panic, ever.), we should go with that one.

bm-w:
Glad to get positive feedback! Sure thing, I’ll refactor and use an existing `Duration` type.

To peek ahead a little, this is essentially what I’ve done with the `DateTime` type [on a branch I’ve been working on that iterates on #213](https://github.com/bm-w/cookie-rs/commit/a373be5#diff-ed12fac37148ceec7aaba6b3de8995be662d0d106cdae22a287cc0ad34354749R141), where I’m either using `time::OffsetDateTime` or `chrono::DateTime<chrono::FixedOffset>` depending on which feature is enabled, and using the former if both are enabled (in that case the `chrono` feature basically only provides some conversions for convenience):
```
#[cfg(any(feature = "time", feature = "chrono"))]
pub mod expiration {
    pub struct DateTime(
        #[cfg(feature = "time)]
        time::OffsetDateTime,
        #[cfg(all(feature = "chrono", not(feature = "time")))]
        chrono::DateTime<chrono::FixedOffset>,
    );
}
```

If it’s OK with you, I’ll do something similar with `Duration` — use the type from either `time` or `chrono` if either or both of those features are enabled, and otherwise fall back to `std::time::Duration` so that at least the `Max-Age` parameter can still be used:
```
pub mod max_age {
    pub struct Duration(
        #[cfg(not(any(feature = "time", feature = "chrono"))]
        std::time::Duration,
        #[cfg(feature = "time")]
        time::Duration,
        #[cfg(all(feature = "chrono", not(feature = "time")))]
        chrono::Duration,
    );
}
```

SergioBenitez:
> To peek ahead a little, this is essentially what I’ve done with the `DateTime` type [on a branch I’ve been working on that iterates on #213](https://github.com/bm-w/cookie-rs/commit/a373be5#diff-ed12fac37148ceec7aaba6b3de8995be662d0d106cdae22a287cc0ad34354749R141), where I’m either using `time::OffsetDateTime` or `chrono::DateTime<chrono::FixedOffset>` depending on which feature is enabled, and using the former if both are enabled (in that case the `chrono` feature basically only provides some conversions for convenience):

Nice! This is pretty neat.

My suggestion is similar, with the caveat that I was suggesting having _no_ wrapper methods of any kind. Instead, just re-export `time` and/or `chrono` and provide _only_ the `From` impls. This is fairly similar to what we have today, except of course that today there is no `chrono` option. 

The other difference is that I was suggesting always using one of the two, and converting to the other. In other words, choosing one of them to be our "native" representation and the other as a foreign conversion: the struct would always contain the native type, irrespective of enabled features, and we'd convert as needed. The advantage here is a much simpler and less error-prone implementation: we don't have to write and test functionality we use internally twice. The disadvantage is that if a consumer enables only the foreign feature, both crates will be compiled in.

I think we can take a hybrid approach that largely negates the disadvantages I present above:

```rust
enum DateTime {
    #[cfg(feature = "chrono")]
    Chrono(chrono::DateTime<Utc>),
    #[cfg(feature = "time")]
    Time(time::OffsetDateTime),
}
```

This way, there's never any conversion from one type to the other necessary. We still need to write the functionality we use internally twice, but even then we should have more clarity.

To summarize:
  * Use an `enum` as above.
  * Distill internal usage of date/time in `cookie` into a private API that multiplexes based on the enabled features. For example, the `parse_date()`  function in `parse.rs` should become a method of `DateTime`, with the four calls in `parse_inner()` rolled into that method, implemented twice, and then called from `parse_inner()`.
  * For tests, choose whichever, and then just `cfg` gate the test appropriately. We should never see any `cfg!()` related to time outside of this internal `time` module.
  * The only public API should be via `From` and `Into` impls, and perhaps `PartialEq`, `Eq`, `Hash`, etc. as needed. Everything else should be done by the client via the native `time` and `chrono` types.

Using this outside of `cookie` would look like:

```rust
use cookie::time;
use cookie::chrono;

let a = Cookie::build("foo", "bar").expires(time::OffsetDateTime::now_utc());
let b = Cookie::build("foo", "bar").expires(chrono::Utc::now());
assert_eq!(a, b);
```

SergioBenitez:
Also, you can use this trick to group all the of the respective `impl`s:

```rust
#[cfg(feature = "chrono")]
mod chrono_impls {
    use super::*;

    impl From<chrono::Foo> for DateTime { .. }

    impl From<DateTime> for chrono::Foo {
        fn from(v: DateTime) -> chrono::Foo {
            match v {
                DateTime::Chrono(v) => v,
                #[cfg(feature = "time")]
                DateTime::Time(v) => chrono::Foo::from(v.timestamp()).utc(),
            }
        }
    }

    impl PartialEq<chrono::Foo> for DateTime { .. }
    impl PartialOrd<chrono::Foo> for DateTime { .. }

    fn parse_cookie_date(..) -> Result<Self> { .. }
}

#[cfg(feature = "time")]
mod time_impls {
    use super::*;

    impl From<time::Foo> for DateTime { .. }
    impl From<DateTime> for time::Foo { .. }
    impl PartialEq<time::Foo> for DateTime { .. }
    impl PartialOrd<time::Foo> for DateTime { .. }

    fn parse_cookie_date(..) -> Result<Self> { .. }
}

#[cfg(any(feature = "time", feature = "chrono"))]
impl DateTime {
    pub(crate) fn parse_cookie_date(string: &str) -> Result<Self> {
        #[cfg(all(feature = "time", not(feature = "chrono")))]
        chrono_impl::parse_cookie_date(string)

        #[cfg(all(feature = "chrono", not(feature = "time")))]
        time_impl::parse_cookie_date(string)

        #[cfg(all(feature = "chrono", feature = "time"))]
        chrono_impl::parse_cookie_date(string)
    }
}
```

bm-w:
Was coding up those last two commits while you made that last comment, causing a bit of cross-talk there… what you’re suggesting sounds good after a quick read! I’ll dig deeper once I can get back to this.

> if a consumer enables only the foreign feature, both crates will be compiled in.

This is the one thing I wanted to avoid, but I think that’s achievable using your `enum`.

bm-w:
I’ve pushed more changes both to this PR and to the [`chrono-additive`](https://github.com/SergioBenitez/cookie-rs/compare/master...31ecfa5) branch. Right now I’m pretty happy with where I’ve been able to take things, but I’m hitting one final hiccup: I can’t get the `Cookie::build("foo", "bar").expires(time/chrono::…)` syntax to work.

The problem is that there is already that `impl<T: Into<Option<DateTime>>> From<T> for Expiration { … }` that conflicts with any of my attempts to do something like `impl<T: Into<DateTime>> From<T> for Expiration { … }`. I can see two ways around it: either remove that former `impl` (and require `.set_expires(Expiration::Session)` instead of `.set_expires(None)`), or add `set_expires_datetime<T: Into<DateTime>>(…)` methods (to `Cookie` and `CookieBuilder`. Or maybe you see another way?

By the way, you’ll note that I haven’t yet done much of the refactoring you suggested (the `mod time/chrono_impl { … }`, or the `enum DateTime { … }`, etc.). I don’t think it’ll help us with the mentioned hiccup, and if we do resolve that the result is essentially working. Part of me feels the refactoring might be a lot of work for little gain (i.e. just cosmetic code changes). Please let me know how important it would be to you (and if you might be willing to take it on yourself?).

SergioBenitez:
I spent some time coming up with a "scaffold" of sorts for how I think this should be implemented. Please check out the `time-chrono` branch.

SergioBenitez:
Closing due to inactivity.

PR #222, opened 2024-01-30
feat(builder): add (set) value method

MrAntix:
this allows for builder patterns where the value is, or is not, set

``` mock.set_cookie("NAME", |b| b.value("VALUE").path("/")); ```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": false
+}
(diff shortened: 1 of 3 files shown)


SergioBenitez:
Please remove all changes with respect to formatting. I'll take a look as soon as that's removed.

MrAntix:
Ah, sorry, thanks for the quick check, I should have noticed

I have taken the liberty of adding a settings file for VS Code which will stop this happening again, should people have auto format switched on - is that ok?

Ant

PR #230, opened 2024-10-26
fix: year check

pixel365:
RFC 6265, Page 15

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -9,6 +9,7 @@ use std::ascii::AsciiExt;
 
 #[cfg(feature = "percent-encode")]
 use percent_encoding::percent_decode;
+use time::error::InvalidVariant;
 use time::{PrimitiveDateTime, Duration, OffsetDateTime};
 use time::{parsing::Parsable, macros::format_description, format_description::FormatItem};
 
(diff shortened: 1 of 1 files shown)


SergioBenitez:
I don't think we should do this, at least not this way, as doing so would mean we fail to parse the entire cookie if the date is < 1601, which would violate the spec (only the expires / date field should fail to parse).

The question of whether the date should or shouldn't parse if the year is < 1601 is a tricky one. Obviously according to the spec, it should fail to parse. Chrome, however, [allows it and returns a "minimal time"](https://github.com/chromium/chromium/blob/92076384fdb9a9147b7c0e5bede0ccff1dc3558b/net/cookies/cookie_util_unittest.cc#L470-L497). Firefox [accepts it](https://github.com/pcwalton/mozilla-central/blob/b4e59f707a52cc4abca64b4c3356a240349ac2c1/nsprpub/pr/src/misc/prtime.c#L1596-L1661). The [`jshttp/cookie`](https://github.com/jshttp/cookie) library also parses it just fine (using the browser's `Date`). In short, we'd be doing something quite different from the norm, and I suspect we'd break a few things.

The fix to offset is correct, however.

SergioBenitez:
Correct fixes merged in https://github.com/rwf2/cookie-rs/commit/ff615ab285b0b01843ceb22f5b5d98001d8fd7a3.

Issue #227, opened 2024-10-06
`chrono` support


musjj:
It looks like that this library requires `time-rs/time`, would it be possible to support `chrono` too? I'm using `chrono` in my project and I had to add a `time` dependency just for cookie functionalities. Having to convert between the two types is also pretty awkward and probably error-prone.

algesten:
Notice that these days there is also [jiff](https://crates.io/crates/jiff) which maybe is an even better consideration?

Issue #162, opened 2020-08-11
Parser incorrectly parses years in some cases


except:
Consider the following Set-Cookie header - 
`foo=bar; expires=Thu, 10-Sep-20 20:00:00 GMT;`

This should be parsed as `2020-09-10 20:00 +0` however expires returns `0020-09-10 20:00 +0`

jhpratt:
After a quick discussion, this behavior is correct as intended on time's end — the century will not be assumed.

Apparently this originates from Cloudflare, which isn't actually following the spec (emphasis mine)

>  HTTP/1.1 clients and servers that parse the date value MUST accept
   all three formats (for compatibility with HTTP/1.0), **though they MUST
   only generate the RFC 1123 format for representing HTTP-date values
   in header fields.**

The first part of that doesn't actually specify what _should_ happen when encountering a two-digit year, though I can't find anything along those lines in any RFC. As such, I think it's best to follow browsers' behavior. Apparently Chrome uses the following code (@except can you provide a source?)

```c
// Normalize the year to expand abbreviated years to the full year.
if (exploded.year >= 69 && exploded.year <= 99)
  exploded.year += 1900;
if (exploded.year >= 0 && exploded.year <= 68)
  exploded.year += 2000;
```

This essentially assumes the year is between 1969 and 2068 inclusive.

Unfortunately, the only way to adjust the year is to fully reconstruct a `PrimitiveDateTime` from its components. Having a better API for operations along these lines is tracked in time-rs/time#256.

except:
Source to where [Chromium](https://chromium.googlesource.com/chromium/src/net/+/master/cookies/cookie_util.cc#316) corrects two-digit based years.

jhpratt:
For the record: I have just submitted a request to Cloudflare. Regardless, this crate's behavior is also not spec compliant.

except:
Firefox uses PR_ParseTimeString ([PR_ParseTimeStringToExplodedTime](https://hg.mozilla.org/mozilla-central/file/tip/nsprpub/pr/src/misc/prtime.c#l1469)) which does something similar when [parsing](https://hg.mozilla.org/mozilla-central/file/tip/netwerk/cookie/CookieService.cpp#l1908).

```c++
year = n1;
if (year < 70) {
    year += 2000;
}
else if (year < 100) {
    year += 1900;
}
```

[...]

PR #233, opened 2025-02-21
feature: chrono support

ozpv:
this pr will add `chrono` support to the crate and close issue #227 

So far, I've created type aliases to the respective date/time types in `chrono` and `time` to make the implementation easier.

I found something wrong with this approach through. It works fine to have the drop-in replacement feature flag for new projects or projects that don't already depend on cookie in some way, but existing ones would have to implement support for the `chrono` feature flag to be able to take advantage of this.

If this isn't fine, I could create a seperate module with something like a `ChronoCookie`, which would just be a copy of Cookie using `chrono` instead. Then it would be possible to use `Cookie::into_chrono` or vice versa to convert between. The drawback of this approach is maintaining the conversions between `time` and `chrono`, and also requiring the `time` dependency as a `chrono` user, even if you never actually use it

Currently, however, it compiles with all features and all the parsing tests are passing for the `chrono` feature. I will get the rest of the tests working in the next couple of days and update the test script

A couple of updates outside of this feature would be to commit 3795f2e, where now abbreviated dates before 70 will be 21st century dates. This is because the browsers referenced in issue #162 have [updated year normalization.](https://chromium.googlesource.com/chromium/src/net/+/master/cookies/cookie_util.cc#563)

The other thing I updated was the tests for abbreviated dates, the day of the week was mismatched with the actual day, causing the `chrono` test to fail (it passes now). I'm unsure if it's the custom parsing function for short `time` dates that ignores this, or `time` itself, but that seems like it could be an issue

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,7 +100,7 @@ use std::str::FromStr;
 
 #[allow(unused_imports, deprecated)] use std::ascii::AsciiExt;
 
-#[cfg(feature = "chrono")] use chrono::{DateTime, Utc};
+#[cfg(feature = "chrono")] use chrono::{DateTime, NaiveDate, Utc};
 #[cfg(not(feature = "chrono"))] use time::{macros::datetime, OffsetDateTime, UtcOffset};
 
 pub use crate::builder::CookieBuilder;
@@ -1132,7 +1132,10 @@ impl<'c> Cookie<'c> {
     /// assert_eq!(c.expires(), Some(Expiration::Session));
     /// ```
     pub fn set_expires<T: Into<Expiration>>(&mut self, time: T) {
-        #[cfg(feature = "chrono")] static MAX_DATETIME: UtcDateTime = DateTime::<Utc>::MAX_UTC;
+        #[cfg(feature = "chrono")] 
+        static MAX_DATETIME: DateTime<Utc> 
+            = NaiveDate::from_ymd_opt(9999, 12, 31).unwrap().and_hms_nano_opt(23, 59, 59, 999_999).unwrap().and_utc();
+
         #[cfg(not(feature = "chrono"))] static MAX_DATETIME: OffsetDateTime = datetime!(9999-12-31 23:59:59.999_999 UTC);
 
         // RFC 6265 requires dates not to exceed 9999 years.
@@ -1701,8 +1704,10 @@ impl<'a> AsMut<Cookie<'a>> for Cookie<'a> {
 
 #[cfg(test)]
 mod tests {
-    use crate::{parse::parse_date, Cookie, SameSite};
-    use time::{Duration, OffsetDateTime};
+    use crate::{Cookie, SameSite};
+    #[cfg(feature = "chrono")] use chrono::{Duration, DateTime, NaiveDate, NaiveDateTime};
+    #[cfg(not(feature = "chrono"))] use time::{Duration, OffsetDateTime};
+    #[cfg(not(feature = "chrono"))] use crate::parse::parse_date;
 
     #[test]
     fn format() {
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -94,17 +98,29 @@ use std::borrow::Cow;
 use std::fmt;
 use std::str::FromStr;
 
-#[allow(unused_imports, deprecated)]
-use std::ascii::AsciiExt;
+#[allow(unused_imports, deprecated)] use std::ascii::AsciiExt;
 
-use time::{Duration, OffsetDateTime, UtcOffset, macros::datetime};
+#[cfg(feature = "chrono")] use chrono::{DateTime, NaiveDate, Utc};
+#[cfg(not(feature = "chrono"))] use time::{macros::datetime, OffsetDateTime, UtcOffset};
```
Caellian: Why are these attributes inlined?

src/lib.rs
```diff
@@ -163,13 +179,13 @@ impl<'c> CookieStr<'c> {
                     Cow::Borrowed(s) => Some(&s[i..j]),
                     Cow::Owned(_) => None,
                 }
-            },
+            }
```
Caellian: not needed

Caellian:
Avoid re-parsing the format string at runtime:

```rust
    use chrono::format::Item;
    use chrono::format::{Item::*, Fixed::*, Numeric::*, Pad, Pad::Zero};
    
    // Following were produced by debug printing corresponding format strings using:
    // `chrono::format::StrftimeItems::new(FORMAT_STRING).parse().unwrap()`

    /// Format: %a, %d %b %Y %H:%M:%S GMT
    pub static FMT1: &[Item] = &[Fixed(ShortWeekdayName), Literal(","), Space(" "), Numeric(Day, Zero), Space(" "), Fixed(ShortMonthName), Space(" "), Numeric(Year, Zero), Space(" "), Numeric(Hour, Zero), Literal(":"), Numeric(Minute, Zero), Literal(":"), Numeric(Second, Zero), Space(" "), Literal("GMT")];
    /// Format: %A, %d-%b-%y %H:%M:%S GMT
    pub static FMT2: &[Item] = &[Fixed(LongWeekdayName), Literal(","), Space(" "), Numeric(Day, Zero), Literal("-"), Fixed(ShortMonthName), Literal("-"), Numeric(YearMod100, Zero), Space(" "), Numeric(Hour, Zero), Literal(":"), Numeric(Minute, Zero), Literal(":"), Numeric(Second, Zero), Space(" "), Literal("GMT")];
    /// Format: %a %b %_d %H:%M:%S %Y
    pub static FMT3: &[Item] = &[Fixed(ShortWeekdayName), Space(" "), Fixed(ShortMonthName), Space(" "), Numeric(Day, Space), Space(" "), Numeric(Hour, Zero), Literal(":"), Numeric(Minute, Zero), Literal(":"), Numeric(Second, Zero), Space(" "), Numeric(Year, Zero)];
    /// Format: %a, %d-%b-%Y %H:%M:%S GMT
    pub static FMT4: &[Item] = &[Fixed(ShortWeekdayName), Literal(","), Space(" "), Numeric(Day, Zero), Literal("-"), Fixed(ShortMonthName), Literal("-"), Numeric(Year, Zero), Space(" "), Numeric(Hour, Zero), Literal(":"), Numeric(Minute, Zero), Literal(":"), Numeric(Second, Zero), Space(" "), Literal("GMT")];
```

Also, the diff on this PR is extremely noisy. It's got a refactor, addition of chrono and another separate issue packed into two commits. Looking at the changes makes it very difficult to see what adding chrono actually did to the codebase - which is why it would be better those were in separate PRs.

SergioBenitez:
I'm sorry I never got to this, but please note that the implementation here wouldn't be tenable because the chrono feature isn't additive. Please see existing closed PRs for more rationale.

PR #234, opened 2025-03-14
Update cookie-rs (`rand` v0.9 & edition)

markcda:


Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "cookie"
 version = "0.18.1"
 authors = ["Sergio Benitez <sb@sergio.bz>", "Alex Crichton <alex@alexcrichton.com>"]
-edition = "2018"
+edition = "2024"
```
paolobarbolini, line 5: This is a bit too aggressive. Why bump MSRV to 1.85 when you're not benefiting from it?

SergioBenitez: Agreed! We want to keep the MSRV as low as possible.

SergioBenitez:
Please limit your PR to updating `rand`. Specify as well what the new MSRV would be with this change, ideally by showing a failure to compile with version `x - 1` and a successful compilation with `x`.

JonathanWoollett-Light:
@SergioBenitez I've opened a PR just updating `rand` https://github.com/rwf2/cookie-rs/pull/242

SergioBenitez:
I updated `rand` to 0.10 in https://github.com/rwf2/cookie-rs/commit/ecfd54f456d6a4f05cdab5fc633d33c7edee90bf.

PR #239, opened 2025-04-26
Drop build script

paolobarbolini:
While looking at uses of build script in my dependency tree, I've realized that this crate uses a build script just to turn on the `doc_cfg` nightly feature. Most other crates instead solve this problem by using `rustdoc-args` in the docs.rs config. This switches to that, removing the build script and the dependency on `version_check`.


SergioBenitez:
Merged (in essence) in https://github.com/rwf2/cookie-rs/commit/c21e0bd5a75f9c2e9d37ab5c601da5da9a1f04d7.

PR #246, opened 2025-12-16
Exclude development scripts from published package

weiznich:
During a dependency review we noticed that the cookie crate includes various development scripts. These development scripts shouldn't be there as they might, at some point become problematic. As of now they prevent any downstream user from enabling the `[bans.build.interpreted]` option of cargo deny.

I opted for using an explicit include list instead of an exclude list to prevent these files from being included in the published packages to make sure that everything that's included is an conscious choice.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ description = """
 HTTP cookie parsing and cookie jar management. Supports signed and private
 (encrypted, authenticated) jars.
 """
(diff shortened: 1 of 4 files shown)


SergioBenitez:
Merged a version of this in https://github.com/rwf2/cookie-rs/commit/157e0ea7430b37614913343d991e36e39792b1d9. Thank you!

Issue #255, opened 2026-06-30
Compilation issue due to update in `time`


TobiasDeBruijn:
Due to an update in the `time` crate, since v0.3.52, the cookie crate no longer compiles:
```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
   --> /home/tobias/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cookie-0.16.2/src/parse.rs:263:27
    |
263 |     let mut date = format.parse(s.as_bytes())?;
    |                           ^^^^^-------------- argument #2 of type `Option<Parsed>` is missing
    |
note: method defined here
   --> /home/tobias/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/time-0.3.52/src/parsing/parsable.rs:62:12
    |
 62 |         fn parse(&self, input: &[u8], defaults: Option<Parsed>) -> Result<Parsed, error::Parse> {
    |            ^^^^^
help: provide the argument
    |
263 |     let mut date = format.parse(s.as_bytes(), /* Option<Parsed> */)?;
    |                                             ++++++++++++++++++++++

For more information about this error, try `rustc --explain E0061`.
error: could not compile `cookie` (lib) due to 1 previous error
```

As per the `time` author, this is on Cookie to fix:
> This is due to cookie relying on an internal API. Please report the issue there, as that's where the issue lies. I have never documented the existence of that method, nor have I ever guaranteed anything regarding semver on that front.

[Relevant issue on time's GitHub](https://github.com/time-rs/time/issues/793)

EvanCarroll:
If the author of time's issue is that documentation determines support, that's a pretty flimsy ledge to stand on. Obviously, if you have cookie depending on undocumented features the right course of action here is to yank the offending version of time and push out another major-version release of time which changes

1. the behavior to the new desired version,
4. seals the impl because documentation doesn't determine what is a breaking change (even if only by convention).


I look forward to the version of cookie.rs that gates use of time/chrono behind a feature flag.

neunzehn86:
Yeah, for some reason, that guy is really pissed.
He shuts down any discussion by closing and locking issues.

Powerbyte7:
@neunzehn86 Having every affected person personally chime in is pointless, and exhausting. It'd be different if people were actually submitting PRs to fix the problem, but it's mostly noise. If you're serious about discussing the versioning approach, don't do it in the heat of the issue. Open source is taken for granted enough already.

[...]

PR #256, opened 2026-06-30
Patched the parsing code that uses time and bumped the rust version

tunafysh:
* Patched the code in src/parse.rs from `format.parse(s.as_bytes())` to `format.parse(s.as_bytes(), None)`by providing the value `None` to the `defaults` argument. 
* Also bumped the rust version from 1.85 to 1.88 because time 0.3.52 requires it.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -242,7 +242,7 @@ pub(crate) fn parse_cookie<'c, S>(cow: S, decode: bool) -> Result<Cookie<'c>, Pa
 
 pub(crate) fn parse_date(s: &str, format: &impl Parsable) -> Result<OffsetDateTime, time::Error> {
     // Parse. Handle "abbreviated" dates like Chromium. See cookie#162.
-    let mut date = format.parse(s.as_bytes())?;
+    let mut date = format.parse(s.as_bytes(), None)?;
     if let Some(y) = date.year().or_else(|| date.year_last_two().map(|v| v as i32)) {
         let offset = match y {
             0..=69 => 2000,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "cookie"
 version = "0.18.1"
 authors = ["Sergio Benitez <sb@sergio.bz>", "Alex Crichton <alex@alexcrichton.com>"]
 edition = "2024"
-rust-version = "1.85"
+rust-version = "1.88"
 license = "MIT OR Apache-2.0"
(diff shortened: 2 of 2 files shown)


dns2utf8:
Until this is merged:
```toml
[patch.crates-io]
cookie = { git = "https://github.com/tunafysh/cookie-rs/", rev = "45612d0f23e0a077ab9c78280770ba06037e5437" }
```

msrd0:
I don't think this should be fixed here but rather in the `time` crate. The accidental breaking change has already been reported there: https://github.com/time-rs/time/issues/794

krishna-santosh:
> I don't think this should be fixed here but rather in the `time` crate. The accidental breaking change has already been reported there: [time-rs/time#794](https://github.com/time-rs/time/issues/794)

https://github.com/time-rs/time/issues/793#issuecomment-4841843595

SamRodri:
I think a better fix would be `time::Date::parse(s, format)` because the `Parsable::parse` method is not public API

src/parse.rs
```diff
@@ -242,7 +242,7 @@ pub(crate) fn parse_cookie<'c, S>(cow: S, decode: bool) -> Result<Cookie<'c>, Pa
 
 pub(crate) fn parse_date(s: &str, format: &impl Parsable) -> Result<OffsetDateTime, time::Error> {
     // Parse. Handle "abbreviated" dates like Chromium. See cookie#162.
-    let mut date = format.parse(s.as_bytes())?;
```
dcormier, line 245: The solution probably should not continue to use the undocumented method with no semver guarantees. Relevant comment from the `time` maintainers: https://github.com/time-rs/time/issues/793#issuecomment-4841843595

tunafysh: yeah but this is meant as a temporary replacement as tauri is literally refusing to build.

joshka:
See https://github.com/rwf2/cookie-rs/compare/master...joshka:cookie-rs:joshka/avoid-private-time-parsing for a better approach to this that both fully avoids the problem and adds more testing to show that the parts involved are properly covered. More details in the #255

src/parse.rs
jhpratt: This "temporary replacement" will be broken in the future. It is not the solution you think it is. You will not get an extraordinarily clear deprecation warning.

SergioBenitez:
Closing since this continues to use a private API, which is what caused this issue in the first place.

PR #258, opened 2026-06-30
fix build issue with time parse fn

tronicboy1:
You don't have to use this if you have a better fix, I am just going to use this in my projects to fix our immediate CICD issues. Feel free to close if you have a better solution.


demurgos:
I recommend checking existing PRs when sending a fix: this is a duplicate of #256, with the same problem of still using a private method of `time`.

